PR #71, opened 2015-12-18
package.json: use ${workspaceRoot}

isidorn:
This should only be merged in when the VSCode january update is out due to this issue https://github.com/Microsoft/vscode/issues/1180

At the moment we magically append the workspace root to some properties of the launch.json on the vscode side.
This will go away in the future since it is clunky. Thus we want to educate users to use the ${workspaceRoot} variable for all paths, since we will no longer support relative paths in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
             "name": "Launch index.html",
             "type": "chrome",
             "request": "launch",
-            "file": "index.html"
+            "file": "${workspaceRoot}/index.html"
           },
           {
             "name": "Launch localhost with sourcemaps",


roblourens:
Hm, I don't see it as magic or clunky - you're just resolving a relative path from the workspace which totally makes sense to me. On the other hand it seems like most of the users of this extension are using 'url' anyway. Thanks for the PR!

roblourens:
I guess the magic is that it's only certain properties that work that way. And an extension could technically resolve 'file' itself to keep the same behavior.

isidorn:
Yes the magic is that we only do that for a certain number of files..

PR #90, opened 2016-01-27
Fixing an uncaught exception with misbehaved stack frames

MSLaguana:
Some targets (I encountered this on an iOS emulator) include invalid stack frames which would cause an exception to be thrown, and this would cause issues for VS code. With this fix we replace these stack frames with 'Unknown' and otherwise continue working.

Another alternative would be to exclude these stack frames from the response, though I'm not sure whether that is a better outcome.


roblourens:
Thank you Ms. Laguana!

PR #101, opened 2016-02-12
Add animated GIF to README

auchenberg:


paulirish:
Looks good!

danmosemsft:
This particular gif does not appear to render in IE or Edge, only in Chrome. I'll open a bug.

auchenberg:
Interesting. Renders fine in Blink, Webkit and Gecko.

danmosemsft:
It's because GitHub is serving the gif with  "X-Content-Type-Options": "nosniff"  and Content-Type: application/octet-stream and Chrome is ignoring that. I'm sure there's lots of history here I'm unaware of so I'll ask internally. 
More usefully -- how can we get GitHub to serve this as image/gif, I see it does that with [some others](https://github.com/Flipboard/FLAnimatedImage/blob/master/images/flanimatedimage-demo-player.gif)

danmosemsft:
OK, I've done enough reading of Chrome and Mozilla archives to see that this is something under debate. And since I'm not on the IE/Edge team and don't know much about HTTP I'll let them continue that debate. :)

So how do we make GitHub serve it right?

danmosemsft:
I can't take this any further this weekend and I'm gone next week. Kenneth - tag!

auchenberg:
@danmosemsft Using rawgit.com seems to serve the correct mime type. Fixed in https://github.com/Microsoft/vscode-chrome-debug/commit/085b88c936bbc0d012b37444f1f421308d325b74.

mislav:
I'm glad to see that you worked around your issue. GitHub intentionally doesn't serve various file type with the correct `Content-type` to prevent hotlinking and various XSS vectors. Therefore, hotlinking an image from a git repo, even when it's from the same repo, in a manner like this `Demo` isn't designed to work, and only accidentally works on some browsers due to them not respecting `Content-type`.

PR #104, opened 2016-02-16
Adding the option to specify your custom path for the sourcemap file 

jtarquino:
Adding the option to specify your custom path for the sourcemap file to workaround problems when the extension can't download it automatically.
Adding logging for the http failures in the getUrl utility

### DIFF
--- a/adapter/sourceMaps/sourceMapTransformer.ts
+++ b/adapter/sourceMaps/sourceMapTransformer.ts
@@ -25,7 +25,7 @@ export class SourceMapTransformer implements IDebugTransformer {
     private _webRoot: string;
     private _authoredPathsToMappedBPLines: Map<string, number[]>;
     private _authoredPathsToMappedBPCols: Map<string, number[]>;
-    private _externalSourceMapPath : string;
+    private _externalSourceMapPath: string;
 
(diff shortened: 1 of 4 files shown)


roblourens:
Thanks for the PR! I'm not sure this is the right fix though. I'd rather fix this in a way so that it just works as-is, rather than add a workaround. I think we could download the script via the page (F12 dev tools does this), but I'd need to test it. Also, this will replace the sourceMapURL for **all** scripts, which is ok if you just have one, but not useful for pages with more than one script.

To unblock yourself, you could inline your sourcemaps, which should work just fine. Or maybe you can configure your build process to output an absolute local path for the sourceMapURL.

jtarquino:
Thanks , indeed I would like it to just work as is but wasn't able to figure out a different way so this one did the trick for now.
The other part that was hard to find out was the actual http error code that was returning so I added a small line to write it to the log (consider that one when you try the proper fix)
Also if you have the pointers in how to try to download via the page let me know and I can give it. Shoot later

roblourens:
Logging the http error is definitely a good idea. I'd accept a PR with that one line if you want to get a PR in.

PR #107, opened 2016-02-19
Logging http status and message for getUrl failure

jtarquino:
Logging http status and message for getUrl when statuscode is different than 200 (OK), helps to diagnose problems with the debugger extension


roblourens:
Thanks!

PR #110, opened 2016-02-24
Master

SAD1992:


roblourens:
Thanks, but I'm not going to take a PR that just moves code from one branch to another.

PR #118, opened 2016-03-08
Update debugger defaults to absolute paths

MartinMa:
VS Code 0.10.10 (February 2016) no longer 'fixes' relative launch configuration paths in `launch.json`. This is now the same treatment of paths as for task configurations and thus a more consistent and transparent strategy.


auchenberg:
LGTM

PR #135, opened 2016-03-28
Updating code to work as a dependency in other extension projects

jalissia:
For IOS webkit protocol debugging, it is very useful to share this debugger code in a separate extension without needing to duplicate all the code that handles the protocol.
These changes build the required d.ts files into a lib folder and add a couple of fixes to allow other types of projects to extend the core classes and add new webkit protocol targets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 node_modules/
 out/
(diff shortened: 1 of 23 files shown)


roblourens:
Thanks James!

PR #196, opened 2016-07-10
Fix SourceMapPathOverrides in README

ulitiy:


roblourens:
Took this from another PR, but thanks!

PR #241, opened 2016-10-05
Use npm scripts instead of gulp

marvinhagemeister:
This has the nice side-effect of having shorter install times for devs. There is one thing though that I don't understand yet: Although the `sourceMap` option is set to true in `tsconfig.json` the compiler won't emit them (doesn't matter if it's run directly or via gulp).


roblourens:
Sure why not, I don't think gulp is really needed here. I'll pull this in when I get a second.

marvinhagemeister:
Awesome, thank you so much!

roblourens:
Thanks for the PR!

PR #243, opened 2016-10-06
Use new breakpoint syntax

octref:
https://code.visualstudio.com/docs/extensions/example-debuggers#_anatomy-of-the-debug-adapter-packagejson
https://github.com/Microsoft/vscode/issues/9037#issuecomment-233904341


roblourens:
Thanks!

PR #324, opened 2017-01-05
add path mapping

llgcode:
Hi @roblourens,
it fixes Microsoft/vscode-chrome-debug-core#146.
I added pathMapping in the launcher config:
So it replace webRoot and sourceMapPathOverrides that are no longer needed to configure a launcher
I've tested with some typescript and sourcemap and it seems to work fine with just a pathMapping.

It was tricky to configure the debug environment but I finally installed projects and found the right procedure for launching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,10 @@ An example `launch.json` config.
             "type": "chrome",
             "request": "attach",
             "port": 9222,
-            "url": "<url of the open browser tab to connect to>"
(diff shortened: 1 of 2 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

roblourens:
Thanks, will update with this when I have a chance.

PR #392, opened 2017-04-01
Use double quotes consistently in npm scripts.

mlewand:
While I've been trying to debug my configuration I found out that test won't work on Windows machines.

The reason was quoting in package.json. When having apostrophe/single quote used in Windows it passes it directly to the commandline, ending up with wrong arguments.

Repro steps:

1. Clone repo on Windows machine.
1. Open cmd.exe command line in cloned directory.
1. Install deps.
1. Build js files.
1. Run `npm test`.

All tests pass.

No test files are found.


roblourens:
Thanks for the PR!

PR #465, opened 2017-07-25
Added the extraCRDPChannelPort parameter and updated logs to be generated on %tmpdir%\vscode-chrome-debug.txt

digeff:
Added the extraCRDPChannelPort parameter and updated logs to be generated on %tmpdir%\vscode-chrome-debug_%timestamp%.txt

### DIFF
--- a/src/chromeDebug.ts
+++ b/src/chromeDebug.ts
@@ -4,6 +4,8 @@
 
 import {ChromeDebugSession, logger, UrlPathTransformer, BaseSourceMapTransformer} from 'vscode-chrome-debug-core';
 import * as path from 'path';
+import * as os from 'os';
+import * as fs from 'fs';
 import {targetFilter} from './utils';
 
 import {ChromeDebugAdapter} from './chromeDebugAdapter';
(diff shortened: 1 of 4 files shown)


digeff:
@roblourens This changes need the vscode-chrome-debug-core dependency to be updated to the latest version before this can work... Do you want me to change that too?

src/chromeDebug.ts
```diff
@@ -14,9 +16,10 @@ const EXTENSION_NAME = 'debugger-for-chrome';
 declare let VERSION: string;
 let versionWithDefault = typeof VERSION === 'undefined' ? 'unspecified' : VERSION; // Not built with webpack for tests
 
-// non-.txt file types can't be uploaded to github
-// also note that __dirname here is out/
-const logFilePath = path.resolve(__dirname, '../vscode-chrome-debug.txt');
+// non-.txt file types can't be uploaded to github. Generated filename is something like: vscode-chrome-debug_2017-07-19_161210.txt
+const timestampForFilename = new Date().toISOString().replace(/T/, '_').replace(/:/g, '').replace(/\..*/, '');
```
roblourens: Is this only needed for your tests? Could the adapter process be started with an environment var giving the name of the log file? Is that reasonable?

This is fine and tmpdir is fine, but I'm still thinking of a way to keep it writing to the same file on each run for vscode.

How will you find the log file? Are you just looking for the newest one in the tmpdir?

digeff: I only added the timestamp to avoid losing the logs the second time we run the debugger. Would appending to the log file rather than starting from scratch work for you?

roblourens: No, then it would just get huge. I just prefer overwriting one file but we can keep this if it's what you need.

Technically we could specify this in the launch config, because the file name isn't actually used until the logger is initialized. It's just bad design on my part to set up the hypothetical file name ahead of time. But I don't care enough to change it right now.

roblourens:
Can you also update vscode-chrome-debug-core?

src/chromeDebug.ts
digeff: Ok. I'll leave the tmpdir change in this PR (it's the most important change for us), and we might send you a new PR with other logchanges using a launch.json parameter in the future.

roblourens: I vaguely remember this coming up partly because of your tests running multiple instances of the adapter at once, and trampling each others' log files. If that's the case, you may not want to strip the milliseconds portion of the timestamp, because if two adapters start within the same second, they'll get the same log file name. 

If I'm remembering wrong, ignore it...

roblourens:
I didn't mean you have to revert the timestamp logic for the log file name now - I'll merge this but it's totally ok to keep the timestamp in the name if you want.

digeff:
I don't want to make your life more difficult, so if we need this we'll add it later as an optional launch.json parameter as you suggested.

PR #520, opened 2017-10-05
Support for env and cwd options.

### LINKED ISSUE #286: Enhancement: support "env" and "cwd"
I'm using the chrome debugger with electron; it works great but I can't figure out how to set the NODE_ENV environment variable to "development".  It would be great if it supported the "env" key to set environment variables.

Similarly, the "cwd" when running through the chrome debugger is "C:\Program Files (x86)\Microsoft VS Code" which causes problems when using relative paths expecting the project directory

jpap:
Resolves #286.


roblourens:
Thanks!

PR #552, opened 2017-11-17
Adding tests for break on load

rakatyal:


test/int/testSetup.ts
```diff
@@ -13,20 +13,33 @@ import {DebugClient} from 'vscode-debugadapter-testsupport';
...
+function patchLaunchArgs(launchArgs: any): void {
+    formLaunchArgs(launchArgs);
 }
 
 export const lowercaseDriveLetterDirname = __dirname.charAt(0).toLowerCase() + __dirname.substr(1);
 export const PROJECT_ROOT = path.join(lowercaseDriveLetterDirname, '../../../');
 export const DATA_ROOT = path.join(PROJECT_ROOT, 'testdata/');
 
-export function setup(port?: number) {
+export function setup(port?: number, breakOnLoadStrategy?: string) {
```
roblourens: Instead of having a special property for breakOnLoadStrategy, could you extend it to allow any launch config prop to be overridden? ie make this an object and apply all the props above.

test/int/breakOnLoad.test.ts
```diff
@@ -0,0 +1,426 @@
...
+                    if (event.body.category === 'stdout' && event.body.output === 'Hi\n') {
+                        consoleOutputLogged = true;
+                    }
+                }),
+
+                dc.launch({ url: 'http://localhost:7890/index.html', webRoot: testProjectRoot })
+            ]);
+
+            // Wait for sometime to make sure the page load is completed
+            await new Promise(resolve => setTimeout(resolve, 1500));
```
roblourens: You can get rid of the timeout by making the test pass when the console.log message is received.

test/int/breakOnLoad.test.ts
```diff
@@ -0,0 +1,426 @@
...
+
+import {DebugClient} from 'vscode-debugadapter-testsupport';
+import * as ts from 'vscode-chrome-debug-core-testsupport';
+
+import * as testSetup from './testSetup';
+
+suite('BreakOnLoad', () => {
+    const DATA_ROOT = testSetup.DATA_ROOT;
+
+    suite('Regex Approach', () => {
```
roblourens: Are these tests exactly the same as the Instrument tests? Then you could parameterize them and remove the duplication. You can call a function with "regex" or "instrument" that registers the tests with the right type.

rakatyal: Not exactly. Like Instrument has the suite for a 'webpack project' and regex does not. Do you still think this is a good idea?

roblourens: For the tests that are the same, I'd rather see those factored out instead of duplicated.

rakatyal:
@roblourens: Addressed feedback! Thanks.

Issue #532, opened 2017-10-27
Make it easier to debug with Chrome Canary


auchenberg:
Some users prefer debugging in Chrome Canary, and we should make it easier to use Canary over Chrome stable.

We could add a `version` property that can be set to `stable|canary|nightly`..

Related issues: https://github.com/Microsoft/vscode-chrome-debug/issues/531

xymopen:
IMO defining `runtimeExecutable` in `settings.json` and select one by a tag in `launch.json` is a preferred 
way.
eg.
``` js
// settings.json
...
"debug.chrome.runtimeExecutables": {
  "some_version": "path\\to\\chrome.exe",
  "other_version": "path\\to\\another\\chrome.exe",
}
...
```
``` js
// launch.json
...
"runtimeExecutableTag":  "some_version"  // or "${defaultStable}" or something else
// to select default installion of stable channel
// and also some tag for userDataDir
...
```

xymopen:
Since Chromes are installed on the developers' computers - they may locate in different places; while `launch.json` would be commit to the repository, it shouldn't contain the exact path to the executable, or the master of the project have to specific a path that all developers need to install a version of chrome to.

xymopen:
Plus if anyone want to switch release channels, he or she can use a pre-defined tag to choose it.

auchenberg:
@joshunger We can't as this is a Chrome specific browser debugger, and we have to work within a specific version range of Chromium to make sure the debugger and Chromium is compatible. 

@xymopen I see the scenario, but I don't see how custom Chromium versions are applicable to the vast majority of our developers, who use this debugger to debug in Chrome stable installed in the default browser path. 

We have had requests for debugging in Canary which this issue is about, and for the very few using a custom Chrome build the `runtimeExecutable` property has been sufficient.

connor4312:
There are two mechanisms with the new debugger that became the default in 1.47:

 - Setting the runtimeExecutable to `canary`, which auto discovers and runs the canary version if installed
 - In the next release there will be an option to set the default runtimeExecutable in your user settings https://github.com/microsoft/vscode-js-debug/issues/836#issuecomment-721306046

PR #590, opened 2018-01-26
add default installion location of release channels other than stable

xymopen:
according to [Chrome Release Channels](https://www.chromium.org/getting-involved/dev-channel).


auchenberg:
Thanks @xymopen! I think this should be made configurable as proposed in https://github.com/Microsoft/vscode-chrome-debug/issues/532

PR #599, opened 2018-02-13
Changing overlay text based on clients

rakatyal:


roblourens:
And prefix the param with a `_` - we will do this for any new "private" launch config params

rakatyal:
@roblourens: Addressed.

roblourens:
Please don't add it to the package.json, that's what makes it show up for users in launch.json

rakatyal:
Done. Thanks!

PR #601, opened 2018-02-15
Added new integration test

digeff:
Added new integration test testing hitting a breakpoint on the first line of two different script files.


testdata/breakOnLoad_firstLineBp/index.html
```diff
@@ -0,0 +1,10 @@
+<!doctype html>
```
roblourens: Do you need to add new testdata scripts for this? Use an existing script if possible. Otherwise name this for the general scenario that the test script covers, not for the specific test that it was added for.

digeff: Merged with an existing scenario

PR #604, opened 2018-02-27
Now we report the times of the different tasks that happen until we launch and show the user page

digeff:
We add a lot of telemetry to track where the time is going when we launch chrome

Requires this PR: https://github.com/Microsoft/vscode-chrome-debug-core/pull/294/files to be merged first.


digeff:
@roblourens @changsi-an

PR #610, opened 2018-03-08
Now we report the times of the different tasks that happen until we launch and show the user page

digeff:
Now we report the times of the different tasks that happen until we launch and show the user page

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -145,8 +145,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
     public async configurationDone(): Promise<void> {
         if (this._userRequestedUrl) {
             // This means all the setBreakpoints requests have been completed. So we can navigate to the original file/url.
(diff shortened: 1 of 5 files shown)


src/chromeDebugAdapter.ts
```diff
@@ -130,13 +130,26 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+            // Chrome started to navigate to the user's requested url
+            this.events.emit(ChromeDebugSession.FinishedStartingUpEventName);
+        }
+    }
+
+    public async configurationDone(): Promise<void> {
         if (this._userRequestedUrl) {
             // This means all the setBreakpoints requests have been completed. So we can navigate to the original file/url.
             this.chrome.Page.navigate({ url: this._userRequestedUrl });
+            this.events.emitMilestoneReached("RequestedNavigateToUserPage");
```
roblourens: Wait on the navigate before sending the event? Maybe don't await so it doesn't block configurationDone. But just like `Page.navigate(...).then(() => emitMilestoneReached())`

PR #613, opened 2018-03-09
Use landing page even if breakonload is disabled

rakatyal:
As Diego mentioned in the comment, this is required to avoid the race condition during redirections inside VS:
Some web-pages redirect very fast to other web-pages etc... You go to localhost:1234/index.htm and you immediately get redirected to localhost:1234/login.htm
When that happens, chrome-core won't be able to find the url in the list of targets supplied by chrome, and it will fail... If we use the landing page we avoid this race condition

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -101,12 +101,11 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
             }
 
             if (launchUrl) {
-                if (this.breakOnLoadActive) {
-                    // We store the launch file/url provided and temporarily launch and attach to about:blank page. Once we receive configurationDone() event, we redirect the page to this file/url
-                    // This is done to facilitate hitting breakpoints on load
-                    this._userRequestedUrl = launchUrl;
-                    launchUrl = "about:blank";
-                }
+
+                // We store the launch file/url provided and temporarily launch and attach to about:blank page. Once we receive configurationDone() event, we redirect the page to this file/url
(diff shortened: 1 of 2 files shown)


auchenberg:
Why? What's the scenario that requires this?

digeff:
@auchenberg Some web-pages redirect very fast to other web-pages etc... You go to localhost:1234/index.htm and you immediately get redirected to localhost:1234/login.htm
When that happens, chrome-core won't be able to find the url in the list of targets supplied by chrome, and it will fail... If we use the landing page we avoid this race condition

roblourens:
Even though we just talked about this offline but please include some description and context when filing PRs here, this is an open source project and I want to document things publicly. Also I trust github more than I trust SfB to not lose old conversations :)

roblourens:
When breakOnLoad is enabled, the target page blinks and shows the "Paused in..." message briefly when starting up. I thought we had something for it to not show that overlay initially?

That makes me not want to merge this yet, users would see that blink in every scenario.

rakatyal:
@roblourens: You are right but how will this PR impact that scenario? We won't blink if user hasn't set any breakOnLoadStrategy with this change. We will add a landing page but that itself won't cause any breakOnLoad breakpoints correct?
We did make a change which would solve that issue for some of the cases when using "regex" strategy but we didn't make any change for 'instrument' strategy.

roblourens:
Oh right, it won't set up the breakOnLoad BP, just do a very quick navigate.

PR #619, opened 2018-03-16
Launch Chrome in unelevated state on Windows platform by using `wmic call create`.

changsi-an:
This change is needed because when VS is started in elevated mode, DA will be started in elevated mode too. Any processes launched by DA, in our case Chrome, will be started in elevated mode as well. Chrome is known to have issues with elevated mode. 

When DA received `shouldLaunchChromeUnelevated` launch parameter, it will call `cscript.exe` to run a customized Windows Script Host program (`launchUnelevated.js`) to launch chrome process in unelevated mode. 

`launchUnelevated.js` will dump the process id information to a temporary file negotiated with DA, so after the WSH process is finished, DA will poll the content of the file and find out the PID of the launched Chrome process.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -38,7 +38,8 @@ const watchedSources = [
 ];
 
 const scripts = [
-    'src/terminateProcess.sh'
+    'src/terminateProcess.sh',
+    'src/launchUnelevated.js'
 ];
 
 const lintSources = [
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -6,7 +6,7 @@ import * as os from 'os';
 import * as fs from 'fs';
 import * as path from 'path';
 
-import {ChromeDebugAdapter as CoreDebugAdapter, logger, utils as coreUtils, ISourceMapPathOverrides, ChromeDebugSession} from 'vscode-chrome-debug-core';
+import {ChromeDebugAdapter as CoreDebugAdapter, logger, utils as coreUtils, ISourceMapPathOverrides, ChromeDebugSession, telemetry} from 'vscode-chrome-debug-core';
 import {spawn, ChildProcess, fork, execSync} from 'child_process';
 import {Crdp} from 'vscode-chrome-debug-core';
 import {DebugProtocol} from 'vscode-debugprotocol';
@@ -53,7 +53,7 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
             args.breakOnLoadStrategy = 'instrument';
         }
 
-        return super.launch(args).then(() => {
+        return super.launch(args).then(async () => {
             let runtimeExecutable: string;
             if (args.runtimeExecutable) {
                 const re = findExecutable(args.runtimeExecutable);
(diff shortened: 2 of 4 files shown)


gulpfile.js
```diff
@@ -38,7 +38,8 @@ const watchedSources = [
 ];
 
 const scripts = [
-    'src/terminateProcess.sh'
+    'src/terminateProcess.sh',
```
changsi-an: Is "teminatePRocess.sh" still needed?

roblourens: Yes as far as I know. It's mac and Linux only

gulpfile.js
```diff
@@ -38,7 +38,8 @@ const watchedSources = [
...
+    'src/launchUnelevated.js'
```
roblourens: I think this will have to be added to the csproj for signing

src/chromeDebugAdapter.ts
```diff
@@ -222,7 +225,7 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         // Disconnect before killing Chrome, because running "taskkill" when it's paused sometimes doesn't kill it
         super.disconnect(args);
 
-        if (this._chromeProc && !hadTerminated) {
+        if ( (this._chromeProc || (!this._chromeProc && this._chromePID)) && !hadTerminated) {
```
roblourens: Shouldn't happen with the current code, but since this can go through when this._chromeProc is null, add a null check below where we call .kill on it.

src/chromeDebugAdapter.ts
```diff
@@ -403,3 +431,42 @@ function findExecutable(program: string): string | undefined {
...
+    const error = new Error(`Cannot acquire Chrome process id`);
+    let telemetryProperties: any = {
+        semaphoreFileContent: lastAccessFileContent
+    };
+
+    coreUtils.fillErrorDetails(telemetryProperties, error);
+
+    telemetry.telemetry.reportEvent('error', telemetryProperties);
+
+    return null;
```
roblourens: Is this supposed to just keep going and try to attach if it can't find the Chrome pid? Or should it throw and fail the launch?

src/launchUnelevated.js
```diff
@@ -0,0 +1,28 @@
+var objShell = new ActiveXObject("shell.application");
```
roblourens: I'm just going to trust that this is all correct :)

roblourens: Maybe use `let` instead of var.

gulpfile.js
changsi-an: I can't find `src/terminateProcess.sh` in the code base. How does it get injected for max and Linux cases?

gulpfile.js
changsi-an: Looking at the current state of the csproj, when it uses wildcard of `..\out\src\**\*` and `$(OutDir)\**\*.js`, it should include this file too.

gulpfile.js
roblourens: It's here: https://github.com/Microsoft/vscode-node-debug2/blob/master/src/terminateProcess.sh#L1

src/chromeDebugAdapter.ts
changsi-an: Yes, after evaluated two approaches, we reached to a conclusion that we should let the debugging session go as much as it can. Not being able to find process id doesn't necessary mean Chrome is not launched nor non-debuggable. One known side-effect is merely that Chrome can't be closed at the end of the session. Due to current multi-debugger-client design of Chrome, that is not an issue any more.

src/launchUnelevated.js
changsi-an: Yes, and I tested it as much as I can. It works.

Unfortunately the Windows Script Host (cscript.exe) is nowhere near any modern ECMAScript standard. `let` is not supported.

src/chromeDebugAdapter.ts
changsi-an: Good catch. 

Done.

changsi-an:
I am in the process of fixing the current build error. Potentially submit another PR to fix the build error first. So please don't merge this PR until the travis build are all green. :)

PR #622, opened 2018-03-16
Changes to the msbuild signing project to remove unwanted files

msft-cwells:
Removes the dll, pdb, and zip files, which we don't want in the published NuGet package.


msbuild/ChromeDebugAdapter.csproj
```diff
@@ -25,18 +25,26 @@
...
+    <Delete Files="$(OutDir)ChromeDebugAdapter.dll;
+                   $(OutDir)ChromeDebugAdapter.pdb;
+                   $(OutDir)ChromeDebugAdapter.deps.json;
+                   $(OutDir)node_modules\gulp-vinyl-zip\test\assets\archive.zip;
+                   $(OutDir)node_modules\globule\test\fixtures\expand\js\foo.js;
+                   $(OutDir)node_modules\globule\test\fixtures\expand\js\bar.js;
+                   $(OutDir)node_modules\resolve\test\resolver\mug.js;
+                   $(OutDir)node_modules\resolve\test\pathfilter\deep_ref\main.js;
+                   $(OutDir)node_modules\resolve\test\resolver\baz\doom.js;
+                   $(OutDir)node_modules\vsce\node_modules\tmp\test\symlinkme\file.js;
```
roblourens: I'm curious where this list of node_modules comes from?

msft-cwells: That's really the list of files that cause the signing step to fail...but interestingly, they cause the signing step to fail because they are empty JS files (and apparently the signing tool refuses to append our Authenticode cert block to an empty *.js file).

I think we've found a way to simplify some of this process, so hopefully we'll be able to remove that exclusion list in the near future.

roblourens: These all appear to be devDependencies, which don't need to be shipped. I would recommend not shipping any devDependencies, it will make the released version much smaller!

msft-cwells: Indeed.  One of the effects of our planned simplification change will be to not copy/sign/package/ship devDependencies.  Thanks Rob!

PR #623, opened 2018-03-17
Provide target types filter configuration

Yavanosta:
There is no any reasons to prohibit debug of anything but 'page'. For example i'm using this extension for debugging chrome itself, and some internal pages have type equals to 'other'.

Another case is debugging shared or service workers.

Suggestion is to add new configuration parameter for type selection, witch is defaults to `['page']` to save previous behavior, but can be changed to allow debugging other types of content.


Yavanosta:
Should I provide some additional info for this PR, make some changes or do something else?

Yavanosta:
It is ok to debug at least pages with other target types.

JFYI: here is code that determine target type in chromium:
https://chromium.googlesource.com/chromium/src/+/master/chrome/browser/devtools/chrome_devtools_manager_delegate.cc#133

src/chromeDebug.ts
```diff
@@ -18,7 +18,7 @@ ChromeDebugSession.run(ChromeDebugSession.getSession(
         adapter: ChromeDebugAdapter,
         extensionName: EXTENSION_NAME,
         logFilePath: path.resolve(os.tmpdir(), 'vscode-chrome-debug.txt'),
-        targetFilter,
+        targetFilter: targetFilterProvider(),
```
roblourens: This doesn't work. I guess I didn't review it closely enough. Did you test this PR? This also needs changes in the -core library to allow a targetFilter to be determined by the launch config. I'm reverting this PR.

Yavanosta: Oh, I'm really sorry, yes of course i've tested it. Maybe I didn't check all cases. I'll try to rework it and make another PR.

Yavanosta: Could you please help me about changes in core library?

As far as I have seen it need delegate as a filter and does not need any information about what this delegate does. So I thought that generating delegate according to launch config will be enough.

So according to your comment I missed something, could you please point me what?

roblourens: The commented line just generates the default filter for page, since targetFilterProvider doesn't take an argument.

I think that somewhere in -core it will have to read the list of targetFilterTypes and create the filter. That will mean that targetFilter won't be passed to the session here, and ChromeConnection won't take the targetFilter in its constructor. Instead it will be an argument passed to ChromeConnection.attach https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeConnection.ts#L120

Yavanosta: Oh shame on me, i've completely misunderstood purpose of `chromeDebug.ts` file and yes, now I see that I have not test it properly. So i'm sorry for this.

Ok then, we must learn from our mistakes. As far as i see, correct way should look like this:
1. Make PR to `-core` lib which mark  `targetFilter` in [IChromeDebugAdapterOpts](https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugSession.ts#L20)
2. In same PR add support for `targetFilterTypes` in [ICommonRequestArgs](https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/debugAdapterInterfaces.d.ts#L21)
3. Merge this PR into `-core` library if it is possible.
3. Come back here, remove `targetFilter` usage and migrate to `targetFilterTypes`
4. Create Issue in rest of debuggers based on `-core` lib.

Does it sounds like a plan?

roblourens: Yes, that sounds great

PR #630, opened 2018-03-23
Now we provide a reason of why start-up failed

digeff:
When we are asked to go to an unreachable url, we'll add that information to the start-up timings telemetry.

This PR requires this -core PR: https://github.com/Microsoft/vscode-chrome-debug-core/pull/316


roblourens:
Merge conflicts

src/chromeDebugAdapter.ts
```diff
@@ -145,9 +146,17 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
     }
 
     protected onFrameNavigated(params: Crdp.Page.FrameNavigatedEvent): void {
-        if (params.frame.url === this._userRequestedUrl) {
+        const url = params.frame.url;
+        const requestedUrlNoAnchor = this._userRequestedUrl.split('#')[0]; // Frame navigated url doesn't include the anchor
+        if (url === requestedUrlNoAnchor || decodeURI(url) === requestedUrlNoAnchor) { // 'http://localhost:1234/test%20page' will use the not decoded version, 'http://localhost:1234/test page' will use the decoded version
             // Chrome started to navigate to the user's requested url
-            this.events.emit(ChromeDebugSession.FinishedStartingUpEventName);
+            this.events.emit(ChromeDebugSession.FinishedStartingUpEventName, { requestedContentWasDetected: true } as FinishedStartingUpEventArguments);
```
roblourens: Do these all need a cast?

digeff: I use the cast to make sure I'm sending the right properties.
If I don't put the cast, the object is of type any, and anything is considered valid.
If I put the cast, then the compiler will fail if I don't put the valid properties here.

roblourens: I was thinking 'emit' was typed here

roblourens:
Check CI?

roblourens:
Please do this for node2 also

PR #640, opened 2018-04-05
change format 'json' to 'javascript' in code

jonyrock:
If you put 'javascript' in type of code, than it would look much better. No red errors in comments


roblourens:
Good point, thanks

PR #661, opened 2018-05-09
Changes the way we handle process id semaphore file.

changsi-an:
Now we always create a file with randomized file name and delete it after we have read the content of it.

We want to address a risk that when multiple debug adapters are launched by VS simultaneously, each process might overwrite the content of the same file.

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -352,10 +352,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         this.events.emitStepStarted('LaunchTarget.LaunchExe');
         const platform = coreUtils.getPlatform();
         if (platform === coreUtils.Platform.Windows && shouldLaunchUnelevated) {
-            const semaphoreFile = path.join(os.tmpdir(), 'launchedUnelevatedChromeProcess.id');
-            if (fs.existsSync(semaphoreFile)) { // remove the previous semaphoreFile if it exists.
-                fs.unlinkSync(semaphoreFile);
-            }
+            const semaphoreFile = path.join(os.tmpdir(), utils.generateRandomString(8) + '.id');
+
             const chromeProc = fork(getChromeSpawnHelperPath(),
                 [`${process.env.windir}\\System32\\cscript.exe`, path.join(__dirname, 'launchUnelevated.js'),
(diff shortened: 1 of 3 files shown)


src/chromeDebugAdapter.ts
```diff
@@ -534,6 +541,7 @@ async function findNewlyLaunchedChromeProcess(semaphoreFile: string): Promise<st
 
             if (matchedLines.length === 1) {
                 const match = matchedLines[0].match(regexPattern);
+                await cleanUp();
```
digeff: Why do we wait for the cleanUp to finish?

src/chromeDebugAdapter.ts
```diff
@@ -352,10 +352,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         this.events.emitStepStarted('LaunchTarget.LaunchExe');
         const platform = coreUtils.getPlatform();
         if (platform === coreUtils.Platform.Windows && shouldLaunchUnelevated) {
-            const semaphoreFile = path.join(os.tmpdir(), 'launchedUnelevatedChromeProcess.id');
-            if (fs.existsSync(semaphoreFile)) { // remove the previous semaphoreFile if it exists.
-                fs.unlinkSync(semaphoreFile);
-            }
+            const semaphoreFile = path.join(os.tmpdir(), utils.generateRandomString(8) + '.id');
```
digeff: Is it possible that 2 different processes will generate the exact same file path? What will happen in that case?

src/chromeDebugAdapter.ts
```diff
@@ -352,10 +352,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+
```
digeff: How do we handle failures in the launchUnelevated.js file? Do we get any telemetry or logging?

src/chromeDebugAdapter.ts
changsi-an: It's a extreme rare case, even rarer when on the same machine. I don't think we need to consider that.

digeff: Let's add some telemetry for 15.8 just in case this ever happens.

changsi-an:
The PR is closed without being merged, since it is no longer needed.

PR #662, opened 2018-05-10
Added a new package command with a packagePath option

digeff:
We'll use this command in our build system. Generating a .vsix with the same name always makes it easier to configure the build system. We'll use "npm run package -- --packagePath chrome-debug2.vsix"

(Please tell me which version of the code you think is better, this one or the node one, and I'll make both of them match that version)


roblourens:
Also implemented this with minimist

PR #665, opened 2018-05-14
Fix for 'Cannot acquire Chrome process id' error in VS 15.7

digeff:
Line 13 fix is because some users have a - in their username, and the tempFile ended up being C:\Users\some-username\ and wmic failed with invalid path without the quotes

Line 29 fallback is because objShellWindows.count on line 19 is the number of File Explorer windows currently open. If the user doesn't have any File Explorer windows currently open that number is 0, and that logic fails.

In the future we'll remove the objShellWindows.count logic altogether and use the new objShellWindows.FindWindowSW instead.  To try to minimize the chances of breaking anything for this fix we are going to leave the existing way as the default and fall-back to the new way if the old way fails.

### DIFF
--- a/src/launchUnelevated.js
+++ b/src/launchUnelevated.js
@@ -28,7 +28,7 @@ if (objShellWindows != null)
 
     if (objShellWindows.count === 0) {
         // If no File Explorer windows are open fall-back to this alternative way of executing the command
-        var item = objShellWindows.FindWindowSW(0, 0, /*SWC_DESKTOP*/8, 0, /*SWFO_NEEDDISPATCH*/1);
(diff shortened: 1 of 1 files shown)


src/launchUnelevated.js
```diff
@@ -25,4 +25,10 @@ if (objShellWindows != null)
             break;
         }
     }
+
+    if (objShellWindows.count === 0) {
+        // If no File Explorer windows are open fall-back to this alternative way of executing the command
+        var item = objShellWindows.FindWindowSW(0, 0, /*SWC_DESKTOP*/8, 0, /*SWFO_NEEDDISPATCH*/1);
```
changsi-an: How come some of the parameters deserve an annotation, but the rest not?

Maybe this [link](https://msdn.microsoft.com/en-us/library/windows/desktop/cc836568(v=vs.85).aspx) would help adding annotations for all of them.

PR #674, opened 2018-05-31
👌 IMPROVE: Lang identifier for code block

ahmadawais:
Found two code blocks in the documentation without language identifiers. This PR ads those and makes them look good. 👍


ahmadawais:
Glad to help 💜

PR #676, opened 2018-06-01
Call host to launch Chrome unelevated if host sends supportsLaunchUnelevatedProcessRequest flag.

changsi-an:
host sends supportsLaunchUnelevatedProcessRequest flag.

Now PZ host is capable of supporting a `launchUnelevated` request from DA. If supportsLaunchUnelevatedProcessRequest flag is set through `initialize` request, DA will choose to use host's ability to launch Chrome unelevated (when needed).

`spawnChromeUnelevatedWithWindowsScriptHost()` is extracted old code. 

`spawnChromeUnelevatedWithPineZorro()` is the new logic to ask host to launch Chrome unelevated.

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -28,6 +28,10 @@ const DefaultWebSourceMapPathOverrides: ISourceMapPathOverrides = {
     'meteor://💻app/*': '${webRoot}/*'
 };
 
+interface IExtendedInitializeRequestArguments extends DebugProtocol.InitializeRequestArguments{
+    supportsLaunchUnelevatedProcessRequest?: boolean
+}
+
 export class ChromeDebugAdapter extends CoreDebugAdapter {
     private _pagePauseMessage = 'Paused in Visual Studio Code';
 
@@ -35,8 +39,9 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
     private _overlayHelper: utils.DebounceHelper;
     private _chromePID: number;
     private _userRequestedUrl: string;
+    private _doesHostSupportLaunchUnelevatedProcessRequest: boolean;
 
-    public initialize(args: DebugProtocol.InitializeRequestArguments): VSDebugProtocolCapabilities {
+    public initialize(args: IExtendedInitializeRequestArguments): VSDebugProtocolCapabilities {
         this._overlayHelper = new utils.DebounceHelper(/*timeoutMs=*/200);
         const capabilities: VSDebugProtocolCapabilities = super.initialize(args);
         capabilities.supportsRestartRequest = true;
@@ -46,6 +51,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         if (args.locale) {
             localize = nls.config({ locale: args.locale })();
         }
+
+        this._doesHostSupportLaunchUnelevatedProcessRequest = args.supportsLaunchUnelevatedProcessRequest || false;
 
         return capabilities;
     }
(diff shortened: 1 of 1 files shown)


changsi-an:
@digeff @roblourens  @rakatyal

src/chromeDebugAdapter.ts
```diff
@@ -433,6 +431,45 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+            logger.log(`Parsed output file and got Chrome PID ${pidStr}`);
+            return parseInt(pidStr, 10);
+        }
+
+        return null
+    }
+
+    private async spawnChromeUnelevatedWithPineZorro(chromePath: string, chromeArgs: string[]): Promise<number> {
+        return new Promise<number>((resolve, reject) => {
+            this._session.sendRequest("launchUnelevated", {
```
digeff: We are wrapping sendRequest in a promise in a similar way here: https://github.com/Microsoft/vscode-chrome-debug-core/blob/47782a37b6df5efd3a50bf1813019c0068a984fe/src/transformers/fallbackToClientPathTransformer.ts#L36
Would it make sense to extract the async sendRequest logic someplace else?

src/chromeDebugAdapter.ts
```diff
@@ -353,26 +360,17 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
-            });
-
-            const pidStr = await findNewlyLaunchedChromeProcess(semaphoreFile);
+            let chromePid: number;
 
-            if (pidStr) {
-                logger.log(`Parsed output file and got Chrome PID ${pidStr}`);
-                this._chromePID = parseInt(pidStr, 10);
+            if (this._doesHostSupportLaunchUnelevatedProcessRequest) {
+                chromePid = await this.spawnChromeUnelevatedWithPineZorro(chromePath, chromeArgs);
```
digeff: Would this be equivalent to use this._chromePID here instead?

src/chromeDebugAdapter.ts
```diff
@@ -28,15 +28,20 @@ const DefaultWebSourceMapPathOverrides: ISourceMapPathOverrides = {
     'meteor://💻app/*': '${webRoot}/*'
 };
 
+interface IExtendedInitializeRequestArgumetns extends DebugProtocol.InitializeRequestArguments{
```
roblourens: typo

src/chromeDebugAdapter.ts
```diff
@@ -433,6 +431,45 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+
+        if (pidStr) {
+            logger.log(`Parsed output file and got Chrome PID ${pidStr}`);
+            return parseInt(pidStr, 10);
+        }
+
+        return null
+    }
+
+    private async spawnChromeUnelevatedWithPineZorro(chromePath: string, chromeArgs: string[]): Promise<number> {
```
roblourens: Maybe don't name this specifically for pinezorro, just "WithClient"

src/chromeDebugAdapter.ts
roblourens: Does it need telemetry?

changsi-an: @digeff The two usages are not quite the same, they are very different in the parameters they each have respectively. Having a common utility function will carry the different parameters as well, it almost like repeating this code. So extracting a common logic does not improve much, but reduces readability.

@roblourens if the operation does not succeed it will reject, and the call will get an exception thrown then telemetry handler will capture that. Otherwise, there is C# telemetry implemented in the host for each PZ extension requests.

src/chromeDebugAdapter.ts
changsi-an: Fixed.

src/chromeDebugAdapter.ts
changsi-an: Fixed.

src/chromeDebugAdapter.ts
changsi-an: Fixed.

changsi-an:
PTAL

rakatyal:
Can we add tests for this change?

src/chromeDebugAdapter.ts
changsi-an: @dhanvikapila Yes, the admin flag is right on the same DA event (clientRequest/launch). The C# error is on a separate event.

roblourens:
@changsi-an LMK when this is ready to merge. I'm ok with or without tests.

changsi-an:
@roblourens Please help merge it now. I will submit a PR for unit test later.

PR #680, opened 2018-06-07
Add the supportsLaunchUnelevatedProcessRequest flag to the telemetry.

changsi-an:
So we know that VS declared capabilty of launching Chrome unelevated.

Also added a unit test.


changsi-an:
The test and lint issues have been fixed. PTAL.

PR #699, opened 2018-07-09
Updating core and package

rakatyal:


roblourens:
Why do you need to update the version number? I would prefer to only do that when I publish the extension.

rakatyal:
Hey Rob, I removed that. We tend to update the package when we do the VS insertion (for telemetry purposes) which I was doing today but it's not necessary.

PR #703, opened 2018-07-17
Fix race condition when killing Chrome on Windows when disconnecting

mrcrane:


src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,37 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         this._chromeProc = null;
     }
 
+    private async killChromeOnWindows(chromePID: number): Promise<void> {
+        let taskkillCmd = `taskkill /PID ${chromePID}`;
+        logger.log(`Killing Chrome process by pid: ${taskkillCmd}`);
+        try {
+            execSync(taskkillCmd);
+        } catch (e) {
+            // The command will fail if process was not found. This can be safely ignored.
```
digeff: Would it be possible to report other errors via telemetry?

src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,37 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+        }
+
+        for (let i = 0 ; i < 10; i++) {
+            // Check to see if the process is still running
+            let tasklistOutput = execSync(`tasklist /FI "PID eq ${chromePID}"`).toString();
+            if (!tasklistOutput.includes(chromePID.toString())) {
```
digeff: Is it possible that the chromePID is a substring of another process PID?

src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,37 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+        logger.log(`Killing Chrome process by pid: ${taskkillCmd}`);
+        try {
+            execSync(taskkillCmd);
+        } catch (e) {
+            // The command will fail if process was not found. This can be safely ignored.
+        }
+
+        for (let i = 0 ; i < 10; i++) {
+            // Check to see if the process is still running
+            let tasklistOutput = execSync(`tasklist /FI "PID eq ${chromePID}"`).toString();
```
digeff: Can we log this execution too?

src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,37 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
         this._chromeProc = null;
     }
 
+    private async killChromeOnWindows(chromePID: number): Promise<void> {
```
digeff: Do we have enough telemetry to figure it if this method/logic is working as expected or not?

rakatyal:
Why do we have an updated package-lock.json for this change?

src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,37 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+        let taskkillCmd = `taskkill /PID ${chromePID}`;
+        logger.log(`Killing Chrome process by pid: ${taskkillCmd}`);
+        try {
+            execSync(taskkillCmd);
+        } catch (e) {
+            // The command will fail if process was not found. This can be safely ignored.
+        }
+
+        for (let i = 0 ; i < 10; i++) {
```
rakatyal: What's the use for this for loop? How is this better than using a timeout?

rakatyal:
Should we store some time performance telemetry for the disconnect request? PZ gives us 5 seconds to respond to a command and we should make sure we are not near that number after this change.

src/chromeDebugAdapter.ts
mrcrane: When tasklist finds the process it will return output that looks like
`Image Name PID Session Name Session# Mem Usage`
`========== === ============ ======== === =====`
`chrome.exe 16116 Console 1 273,300 K`

When tasklist doesn't find the process it outputs
`INFO: No tasks are running which match the specified criteria.`

In both cases the exit code is 0

Because the tasklist is run with a PID filter it will only return one or zero processes. Any process that with a PID that is a superstring will be filtered out.

I implemented with a search of PID in the output because it seemed like the most robust way to differentiate between success and failure outputs. Searching for the presence of the PID is independent of the format.

Let me know if you have any better suggestions. I can add comments as documentation for what is going on.

src/chromeDebugAdapter.ts
mrcrane: This follows the pattern in `findNewlyLaunchedChromeProcess()` at the bottom of the file. It will try up to 10 times and wait 200ms between tries. Total time will be 2000ms in addition to command execution time. It does use a timeout to wait.

I think it's readable but I am open to changing it.

mrcrane:
The package-lock.json changes happen when I run `npm install`. I assume that is because of a previous check-in. I can leave it out but it keeps coming back so I decided to add it to the PR.

src/chromeDebugAdapter.ts
roblourens: This is fine but you could also use the `retryAsync` helper in core's utils.ts if you want.

src/chromeDebugAdapter.ts
```diff
@@ -347,6 +332,44 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
...
+            // Check to see if the process is still running, with CSV output format
+            let tasklistCmd = `tasklist /FI "PID eq ${chromePID}" /FO CSV`;
+            logger.log(`Looking up process by pid: ${tasklistCmd}`);
+            let tasklistOutput = execSync(tasklistCmd).toString();
+
+            // If the process is found, tasklist will output CSV with one of the values being the PID. Exit code will be 0.
+            // If the process is not found, tasklist will give a generic "not found" message instead. Exit code will also be 0.
+            // If we see an entry in the CSV for the PID, then we can assume the process was found.
+            if (!tasklistOutput.includes(`"${chromePID}"`)) {
```
roblourens: Redundant template string :)

roblourens:
Was the bug here that Chrome takes a bit of time to shutdown, so we are force-killing it unnecessarily?

roblourens:
> I actually can't find any "ClientRequest/disconnect" telemetry in Kusto even though I see all the other requests

Maybe we lose telemetry from right before the process shuts down, let me check that out...

roblourens:
Actually, I think that telemetry after `disconnect` will not be sent (looking at vscode's code at least). The event is sent to the client which then passes it on to the telemetry service, but I think the client will stop handling this telemetry after disconnect.

digeff:
Ship it!

PR #715, opened 2018-08-01
Added new tests to verify webRoot works with any case

digeff:
Added new tests to verify webRoot works with any case


roblourens:
Hey @digeff is this still good to go? I guess I missed it.

digeff:
@roblourens I was never able to run all the integration tests successfully with this change (the test that were failing on my PC were existing tests, not these ones).
Ideally we should work on re-enable the integration tests on PRs, and see if all the tests pass with this change.

PR #728, opened 2018-09-12
Update README

nvh95:
This PR is to update README so others can easily know where to put the config to enable Page refresh feature. It took me an hour to figure it out so I hope other people don't waste the same amount of time as I did.
Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ The supported formats are:
   * A path glob, to skip all scripts that match (like `node_modules/react/*.min.js`)
 
 ## Page refreshing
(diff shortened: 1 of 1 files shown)


roblourens:
I don't think this is the right place but directions on how to edit keybindings, and your steps are mac-specific. But if you update to just point to the docs on that: https://code.visualstudio.com/docs/getstarted/keybindings then I would take this.

nvh95:
Sure @roblourens 
I updated the README. Thanks for your feedback.

PR #731, opened 2018-09-13
Workaround for json/version failing

digeff:
We've seen some cases where json/version fails and we can't get the version. That makes the break-on-load feature not work on the first line on the file on Chrome 69.

We use this workaround to try to use the CDTP Browser.GetVersion API only if the first mechanism fails.

This change depends on this PR: https://github.com/Microsoft/vscode-chrome-debug-core/pull/354

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -17,7 +17,6 @@ import * as errors from './errors';
 
 import * as nls from 'vscode-nls';
 import { FinishedStartingUpEventArguments } from 'vscode-chrome-debug-core/lib/src/executionTimingsReporter';
-import { version } from 'vscode';
 let localize = nls.loadMessageBundle();
 
 // Keep in sync with sourceMapPathOverrides package.json default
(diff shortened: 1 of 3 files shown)


src/chromeDebugAdapter.ts
```diff
@@ -17,6 +17,7 @@ import * as errors from './errors';
 
 import * as nls from 'vscode-nls';
 import { FinishedStartingUpEventArguments } from 'vscode-chrome-debug-core/lib/src/executionTimingsReporter';
+import { version } from 'vscode';
```
roblourens: Is this used? You don't have access to the `vscode` module from here

digeff: Removed

src/chromeDebugAdapter.ts
```diff
@@ -278,6 +278,25 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
              */
             versionInformationPromise.then(versionInformation => telemetry.telemetry.reportEvent('target-version', versionInformation));
 
+            try {
+                if (this._breakOnLoadHelper) {
+                    // This is what -core is doing. We only actually care to see if this fails, to see if we need to apply the workaround
+                    const browserVersion = (await this._chromeConnection.version).browser;
+                    if (browserVersion.isAtLeastVersion(0, 1)) { // If this is false it means it's unknown version
+                        this._breakOnLoadHelper.setBrowserVersion(browserVersion);
```
roblourens: Won't -core do this already?

digeff: That is correct. Removed

PR #739, opened 2018-10-05
Add Indonesian translation

indocomsoft:


roblourens:
Did you translate all of these yourself? Thanks for doing all that work. Our translations come from the Transifex service (https://github.com/Microsoft/Localization/wiki/Visual-Studio-Code-Community-Localization-Project) I can check whether Indonesian can be added as a language. However, most of the code for this extension comes from the https://github.com/Microsoft/vscode-chrome-debug-core repo which has a lot of strings so there would be much more translation work to do...

PR #740, opened 2018-10-05
Uses native `open` executable on macOS to launch Chrome regardless of installation path

### LINKED ISSUE #700: Cannot find Chrome on macOS when not in `/Applications`
- VS Code Version: 1.25

Steps to reproduce:

1. Running macOS; having installed "Google Chrome.app" somewhere else then `/Applications/` (e.g. in `~/Applications/`)
2. Create a launch config like
    ```
    {
      "name": "Debug DevServer",
      "type": "chrome",
      "request": "launch",
      "sourceMaps": true,
      "url": "http://localhost:8080"
    },
    ```
3. -> `[debugger-for-chrome] Error processing "launch"…` … cannot find chrome.

-------

In our development team different members have "Google Chrome.app" installed in different locations. debugger-for-chrome looks for the Chrome only in the `/Applications` folder. I know it's possible to specify another location by configuring a `runtimeExecutable` path. But 1.) other tools (e.g. webpack devServer) are able to do that, and 2.) since the launch config is in our git its a bad place for developer-specific settings.

jamesseanwright:
Fixes #700!

I investigated using [opn](https://github.com/sindresorhus/opn), but this would have resulted in a considerable refactor; I thus opted to lift its logic for invoking Darwin's [`open`](https://scriptingosx.com/2017/02/the-macos-open-command/) executable on macOS. I also determined if a similar change would be required for Linux, but xdg-open doesn't operate in an analogous manner, thus I've left the path intact.

I tested this by moving my Chrome application out of `/Applications` and into another directory; this allowed me to reproduce the issue, and applying the changes in my PR determines that this fix solves the issue.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
             "cwd": "${workspaceFolder}",
             "program": "${workspaceFolder}/out/src/chromeDebug.js",
             "args": [ "--server=4712" ],
-            "outFiles": ["${workspaceFolder}/out/**/*.js"],
+            "outFiles": ["${workspaceFolder}/out/**/*.js"]
         },
         {
             // "debugServer": 4712,
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
             "cwd": "${workspaceFolder}",
             "program": "${workspaceFolder}/out/src/chromeDebug.js",
             "args": [ "--server=4712" ],
-            "outFiles": ["${workspaceFolder}/out/**/*.js"]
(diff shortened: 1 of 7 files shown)


roblourens:
-n does work, but now I'm seeing that when I click the stop button while debugging, it disconnects from the Chrome process but doesn't kill it. I think that it has a handle on the `open` process but not the real Chrome process, and I'm not sure how to get around that.

jamesseanwright:
That's a shame! Leave it with me and I'll see if I can find a clean way of grabbing that handle, otherwise I'll find an alternative.

PR #749, opened 2018-11-01
adding runtime args telemetry for chrome cmd line switches

Shenniey:
This is for obtaining more data in terms of usage of command line switches in chrome


roblourens:
Why do you need this data?

roblourens:
I didn't think VS even had a way to customize this? How does that work?

Shenniey:
The IIS Express dropdown menu has a “browse with” option where you can set what chrome cmd line switches you want to pass to the browser. You can also save the start up configuration for future use.

browse with box
set cmd line switches

roblourens:
Thanks!

PR #792, opened 2019-03-15
Adding initial framework integration tests

EricCornelson:
Adds a handful of integration tests for React, and
some tooling to easily use puppeteer in integration
tests in preparation for v2.


.vscode/launch.json
```diff
@@ -44,6 +44,24 @@
                 "${workspaceFolder}/out/**/*.js"
             ]
         },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "frameworkTest",
+            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+            "args": [
+                "--debug-brk", "--inspect",
```
roblourens: Should not be necessary in a launch config

EricCornelson: Agreed, and removed.

test/int/framework/framework.react.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { setBreakpoint, setConditionalBreakpoint } from '../intTestSupport';
+import { puppeteerSuite, puppeteerTest } from '../puppeteer/puppeteerSuite';
+import { FrameworkTestSuite } from './frameworkCommonTests';
+import { TestProjectSpec } from './frameworkTestSupport';
+
+const DATA_ROOT = testSetup.DATA_ROOT;
+const REACT_PROJECT_ROOT = path.join(DATA_ROOT, 'react');
+const TEST_SPEC = new TestProjectSpec(REACT_PROJECT_ROOT);
+
+puppeteerSuite('React Framework Tests', TEST_SPEC, (suiteContext) => {
```
roblourens: What are the advantages of using puppeteer vs what we already do?

testdata/react/src/App.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react';
```
roblourens: Please include steps to regenerate this project e.g. when we want to use a different version of webpack or react, or make any changes. e.g. other test projects have a tsconfig.json included and I can just run `tsc` to rebuild them.

test/int/framework/framework.react.test.ts
EricCornelson: Basically just making it easier to write tests that interact with the page (e.g. click a button and wait for something to happen).

testdata/react/src/App.js
EricCornelson: Will do. I'm planning on creating a separate repo to house the test project code so as to not weigh down this one with a bunch of react/angular/etc. dependencies

test/int/framework/framework.react.test.ts
roblourens: The goal is just to get some code to run right? For other tests I've just used setInterval or global code. You could even ask the debug adapter to evaluate something which would trigger the code that you want to run.

Also if we use Puppeteer then the debug adapter is no longer launching the process right? That should be tested too.

EricCornelson:
> Can these tests run with v2 at all yet?
Last time I checked they were having some issues, but @digeff is still doing a lot of refactoring so holding off on trying to run them against v2 until he's in a good state.

testdata/react/src/App.js
roblourens: Honestly I think it would be fine to just check in the package.json but not the node_modules. Then we can build/update it in place in case we want to change anything. Seems the simpler than having to find another repo and copy files around?

test/int/framework/framework.react.test.ts
EricCornelson: > The goal is just to get some code to run right? For other tests I've just used setInterval or global code. You could even ask the debug adapter to evaluate something which would trigger the code that you want to run.

Wouldn't doing something like eval have a slightly different impact on things like sourcemaps? Interacting with the DOM closer simulates a real use case. I can check with the team to see if there was a specific scenario they had in mind with puppeteer.

> Also if we use Puppeteer then the debug adapter is no longer launching/managing the chrome process right? That should be tested too.

Nope! The way I've got it set up right now is that we use the debug adapter to launch, and just attach puppeteer to the running instance.

testdata/react/src/App.js
EricCornelson: True. I was just worried about adding a lot of bloat to the repo, but if it works for you I'll just add the whole thing!

EricCornelson: Added the full project under testData/react

roblourens:
Please check out the CI build, I'm not really sure what it's complaining about.

test/int/framework/framework.react.test.ts
roblourens: Oh, I see. It still seems like it could be overkill just to trigger some code? But if you think it will make things easier I think that's ok.

EricCornelson:
Looks like the typings for puppeteer are using conditional types (added in typescript 2.8), and we're on typescript v 2.7.2. Updating to anything above 2.8 resolves the issues. Any reason not to upgrade?

roblourens:
That's fine with me

test/int/framework/framework.react.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        puppeteerTest('Should hit conditional breakpoint in .jsx file', suiteContext, async (context, page) => {
+            let location = { path: TEST_SPEC.src('Counter.jsx'), line: 15};
+            await setConditionalBreakpoint(suiteContext.debugClient, location, 'this.state.count == 2');
+            let incBtn = await page.waitForSelector('#incrementBtn');
+            // click 3 times, state will be = 2 on the third click
+            await incBtn.click();
+            await incBtn.click();
+            incBtn.click();
+            await suiteContext.debugClient.assertStoppedLocation('breakpoint',  location);
+            suiteContext.debugClient.continueRequest();
```
digeff: Should we await this result?

test/int/puppeteer/puppeteerSupport.ts
```diff
@@ -0,0 +1,33 @@
...
+    const browser = await puppeteer.connect({ browserWSEndpoint: webSocketDebuggerUrl, defaultViewport: null });
+    return browser;
+}
+
+/**
+ * Get the first (or only) page loaded in chrome
+ * @param browser Puppeteer browser object
+ */
+export async function firstPage(browser: puppeteer.Browser): Promise<puppeteer.Page> {
+    return browser.pages().then(p => p[0]);
```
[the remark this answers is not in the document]
EricCornelson: Something like this?

```suggestion
    return (await browser.pages())[0];
```

I think I originally had something like that, but thought using `then` was a little more explicit about what's happening (i.e. awaiting the result of `pages()` then returning the first item) .

>If we expect the browser to have a single page, it could make sense to verify this and throw an exception if we find more than one page. That could make it easier to diagnose what is happening if for some reason the browser ever has more than the single page we expect.

Thinking about it a bit more, a better approach is probably just to find the page we want via url, and then throw if we can't find it. It shouldn't matter if there are multiple pages if we can still run our tests correctly.

test/int/intTestSupport.ts
```diff
@@ -0,0 +1,74 @@
...
+        breakpoints: [{ line: location.line, column: location.column, condition }],
+        source: { path: location.path }
+    });
+}
+
+/**
+ * Set a timer. (Note: this method should be used SPARINGLY, if at all, as it will tend to make tests non-deterministic)
+ * @param ms Milliseconds to wait
+ */
+export function timer(ms: number): Promise<any> {
```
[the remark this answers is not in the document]
EricCornelson: Looks like they're both pretty much the same. I'm not using this for anything anymore anyway. I'll take it out.

test/int/framework/frameworkTestSupport.ts
```diff
@@ -0,0 +1,69 @@
...
+    url?: string;
+}
+
+/**
+ * Specifies an integration test project (i.e. a project that will be launched and
+ * attached to in order to test the debug adapter)
+ */
+export class TestProjectSpec {
+
+    _props: ProjectSpecProps;
```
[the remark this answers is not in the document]
EricCornelson: My goal here was to allow some of the properties to be undefined, and then fill them in with sensible defaults, such that you only need to pass props that change from the default (e.g. if 9/10 test projects have the same setup, we don't have to bother passing everything every time, but then still have the freedom to change 1 or 2 things for that 1/10 case).

Definitely open to other suggestions as well.

test/int/framework/framework.react.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        const frameworkTests = new FrameworkTestSuite('React', suiteContext);
+        frameworkTests.testPageReloadBreakpoint({ path: TEST_SPEC.src('App.js'), line: 7 });
+        frameworkTests.testPauseExecution();
+        frameworkTests.testBreakOnLoad({ path: TEST_SPEC.src('App.js'), line: 7 });
+    });
+
+    suite('React specific tests', () => {
+
+        puppeteerTest('Should hit breakpoint in .jsx file', suiteContext, async (context, page) => {
+            let location = { path: TEST_SPEC.src('Counter.jsx'), line: 15};
```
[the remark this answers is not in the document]
EricCornelson: Ah, I like that approach. Using labels would also make it even easier to run the same tests on different project types, just use the same label.

test/int/framework/framework.react.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        });
+
+        puppeteerTest('Should hit conditional breakpoint in .jsx file', suiteContext, async (context, page) => {
+            let location = { path: TEST_SPEC.src('Counter.jsx'), line: 15};
+            await setConditionalBreakpoint(suiteContext.debugClient, location, 'this.state.count == 2');
+            let incBtn = await page.waitForSelector('#incrementBtn');
+            // click 3 times, state will be = 2 on the third click
+            await incBtn.click();
+            await incBtn.click();
+            incBtn.click();
```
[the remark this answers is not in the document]
EricCornelson: Done.

test/int/framework/framework.react.test.ts
EricCornelson: We can actually just remove that last call entirely, as we're done with the test as soon as we verify the stopped location.

test/int/puppeteer/puppeteerSupport.ts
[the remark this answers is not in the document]
digeff: That looks good

test/int/framework/frameworkTestSupport.ts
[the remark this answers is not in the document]
digeff: I'm just suggesting changing the types to reflect that's what you are doing already.
The constructor can have undefined in the parameters, so it uses constructor(props: ProjectSpecProps) which has optional values

in the instance variable prop all the values are defined, because you default to the default values if the user didn't supply it, so all the values will be defined, so the most accurate type would be MakePropertyRequired<ProjectSpecProps>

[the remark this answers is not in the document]
EricCornelson: Got it, I see what you're saying.

test/int/puppeteer/puppeteerSupport.ts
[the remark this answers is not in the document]
EricCornelson: Done

PR #796, opened 2019-03-21
Migrate disconnecting logic to v2. Make int tests kill all chrome instance automatically 

digeff:
I migrated the disconnecting logic to v2 because the intTests were failing because the debugger wasn't killing the instances of Chrome automatically.

After the migration, killing Chrome sometimes works and sometimes doesn't. As a short-term fix I added code so the test would kill all instances of chrome.exe.

TODO: Fix this properly and remove the taskkill for chrome.exe

### DIFF
--- a/test/int/testSetup.ts
+++ b/test/int/testSetup.ts
@@ -7,6 +7,7 @@ import * as path from 'path';
 import * as tmp from 'tmp';
 
 import * as ts from 'vscode-chrome-debug-core-testsupport';
+import { execSync } from 'child_process';
 
 const DEBUG_ADAPTER = './out/src/chromeDebug.js';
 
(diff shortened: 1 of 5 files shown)


test/int/testSetup.ts
```diff
@@ -39,9 +40,12 @@ export function setup(port?: number, launchProps?: any) {
...
-    return ts.setup({entryPoint: DEBUG_ADAPTER, type:'chrome', patchLaunchArgs: patchLaunchArgs, port: port});
+    return ts.setup({ entryPoint: DEBUG_ADAPTER, type: 'chrome', patchLaunchArgs: patchLaunchArgs, port: port });
 }
 
-export function teardown() {
-    return ts.teardown();
+export async function teardown() {
+    // TODO: This is a short-term fix because stopping chrome.exe when the debug session ends is not working properly
+    // We need to fix the product/tests to kill chrome properly and remove this code
+    execSync('taskkill /im chrome.exe', { stdio: 'ignore' });
```
roblourens: Does this throw on non-windows or when chrome.exe doesn't exist?

digeff: It does, nice catch. Surrounding with try/catch...

PR #798, opened 2019-03-25
Run framework tests as part of CI

EricCornelson:
For now, the CI task will only run the framework tests (not all of the integration tests), as these tests are the only ones that can currently run with the embedded chromium instance. Planning on eventually being able to run all integration tests on CI.


EricCornelson:
Checking to see why the tests are failing on mac/linux

digeff:
👍

PR #799, opened 2019-04-01
Merging new framework tests from master into v2

EricCornelson:


roblourens:
Is this just exactly what's in master or are there extra changes?

EricCornelson:
> Is this just exactly what's in master or are there extra changes?

A few small extras, mostly fixing tests.

roblourens:
Fixing tests just for v2? I would like everything to be in master if possible, so is there anything that makes sense to go in master first?

EricCornelson:
I think [4b63f95](https://github.com/Microsoft/vscode-chrome-debug/pull/799/commits/4b63f95e76ae0eebfcbbd9c707864c43e7964c31) is probably relevant to master, otherwise it's just the change to where we're issuing the "kill all instances of chrome" command, and adding bin/obj to the ignore list.

EricCornelson:
> Can you start a PR to bring that into master first then?

Will do.

> 
> > otherwise it's just the change to where we're issuing the "kill all instances of chrome" command
> 
> Why shouldn't that go in master?
> 

That code doesn't exist in master. digeff added it to clean up Chrome instances after tests because v2 is not currently closing Chrome properly. 

> I just want master to be a baseline for all the new tests and I want to make sure we are testing the same thing in both branches.

Agreed. I'm planning on submitting all new test changes to master first, then merging them to v2 separately.

package.json
```diff
@@ -24,6 +24,7 @@
   ],
   "license": "SEE LICENSE IN LICENSE.txt",
   "dependencies": {
+    "mocha-junit-reporter": "^1.18.0",
```
digeff: Should this be a devDependencies?

EricCornelson:
I added [4b63f95](https://github.com/Microsoft/vscode-chrome-debug/pull/799/commits/4b63f95e76ae0eebfcbbd9c707864c43e7964c31) to [this PR ](https://github.com/Microsoft/vscode-chrome-debug/pull/798)

test/int/adapter.test.ts
```diff
@@ -15,12 +15,18 @@ const DATA_ROOT = testSetup.DATA_ROOT;
...
+    let server;
+
     setup(() => {
         return testSetup.setup()
             .then(_dc => dc = _dc);
     });
 
     teardown(() => {
+        if (server) {
+            server.close();
```
digeff: Should we assign null to server in case another tests starts and fails before assigning the server variable to a new server?

test/int/breakpoints.test.ts
```diff
@@ -25,7 +25,7 @@ suite('Breakpoints', () => {
             server.close();
         }
 
-        testSetup.teardown(); // TODO: Should we be awaiting this? This might be the reason that disconnecting is not killing chrome.exe properly
+        await testSetup.teardown(); // TODO: Should we be awaiting this? This might be the reason that disconnecting is not killing chrome.exe properly
```
digeff: You can remove the comment now that you added the await...

package.json
EricCornelson: Yes. Fixed.

test/int/adapter.test.ts
EricCornelson: Fair point. I changed all other instances of teardown to follow the same pattern.

EricCornelson:
I also removed the problematic logging statements in chromeLauncher.ts which were ultimately the root cause of the weird disconnect issue.

test/int/puppeteer/puppeteerSuite.ts
```diff
@@ -0,0 +1,84 @@
...
+  callback: (suiteContext: FrameworkTestContext) => any
+): Mocha.ISuite {
+  return suite(description, () => {
+    let suiteContext: FrameworkTestContext = { testSpec };
+
+    let server: any;
+
+    setup(async () => {
+      suiteContext.debugClient = await testSetup.setup();
+      await suiteContext.debugClient;
```
digeff: Is this await needed?

EricCornelson: Nope, was a remnant of some refactoring. Removed.

PR #811, opened 2019-04-16
Added hit count breakpoints tests and breakpoints utility

digeff:
Added some hit count breakpoints tests
Added a new utility to make it easier to write tests that hit breakpoints

This PR won't compile without these changes: https://github.com/Microsoft/vscode-chrome-debug/pull/810


test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
```diff
@@ -0,0 +1,84 @@
+import * as _ from 'lodash';
```
EricCornelson: Using lodash at all here?

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
```diff
@@ -0,0 +1,84 @@
...
+import { BreakpointsWizard } from '../breakpointsWizard';
+
+export class BreakpointsUpdate {
+    public constructor(
+        public readonly toAdd: BreakpointWizard[],
+        public readonly toRemove: BreakpointWizard[],
+        public readonly toKeepAsIs: BreakpointWizard[]) { }
+}
+
+export interface IPerformChangesImmediatelyOrBatchState {
```
EricCornelson: `IPerformChangesImmediatelyOrBatchState` is really verbose. How about just `IBreakpointsState`.

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
```diff
@@ -0,0 +1,84 @@
...
+    public async batch<T>(batchAction: (fileBreakpointsWizard: FileBreakpointsWizard) => PromiseOrNot<T>): Promise<T> {
+        const batchingUpdates = new BatchingUpdatesState(this, this._state.currentBreakpointsMapping);
+        this._state = batchingUpdates;
+        const result = await batchAction(new FileBreakpointsWizard(this));
+        await batchingUpdates.processBatch(); // processBatch calls sendBreakpointsToClient which will change the state back to PerformChangesImmediatelyState
+        return result;
+    }
+
+    public async sendBreakpointsToClient(update: BreakpointsUpdate): Promise<void> {
+        return new BreakpointsUpdater(this._breakpointsWizard, this, this.client, update, state => this._state = state).update();
```
EricCornelson: Do we need to create a new BreakpointsUpdater on every call?

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    private extractStackTrace(stackTraceResponse: DebugProtocol.StackTraceResponse): string {
+        return stackTraceResponse.body.stackFrames.map(f => this.printStackTraceFrame(f)).join('\n');
+    }
+
+    private printStackTraceFrame(frame: DebugProtocol.StackFrame): string {
+        let frameName = frame.name;
+        if (isThisV1) {
+            // V1 currently has a bug where line numbers are off by 1
+            frameName = frameName.replace(`Line ${frame.line - 1}`, `Line ${frame.line}`);
```
mrcrane: If there is a bug with V1, then the test should fail. We shouldn't be hiding the bug in the tests. If that bug is preventing you from writing this test, then this is a good reason to separate our stack trace tests from the hit count breakpoint tests...

This bug has been fixed in Microsoft/vscode-chrome-debug-core#426 and so it will break this code whenever the new core is published and added to package.json.

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
```diff
@@ -0,0 +1,102 @@
...
+    private validateStackTraceResponse(stackTraceResponse: DebugProtocol.StackTraceResponse) {
+        expect(stackTraceResponse.success, `Expected the response to the stack trace request to be succesful yet it failed: ${JSON.stringify(stackTraceResponse)}`).to.equal(true);
+        expect(stackTraceResponse.body.totalFrames, `The number of stackFrames was different than the value supplied on the totalFrames field`)
+            .to.equal(stackTraceResponse.body.stackFrames.length);
+        stackTraceResponse.body.stackFrames.forEach(frame => {
+            // Warning: We don't currently validate frame.source.path
+            const expectedSourceNameAndLine = ` [${frame.source.name}] Line ${frame.line}`;
+            if (isThisV2) {
+                // V1 has a bug where frame.name is not coherent with frame.line
+                expect(frame.name, 'Expected the formatted name to match the source name and line supplied as individual attributes').to.endsWith(expectedSourceNameAndLine);
```
mrcrane: Same, this can be removed once Microsoft/vscode-chrome-debug-core#426 is in place

test/int/featureBasedSuits/hitCountBreakpointTests.test.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+import * as _ from 'lodash';
+import { puppeteerSuite, puppeteerTest } from '../puppeteer/puppeteerSuite';
+import { reactTestSpecification } from '../resources/resourceProjects';
+import { BreakpointsWizard as BreakpointsWizard } from '../wizards/breakpoints/breakpointsWizard';
+import { asyncRepeatSerially } from '../utils/repeat';
+
+puppeteerSuite('Hit count breakpoints on a React project', reactTestSpecification, (suiteContext) => {
+    suite('Some basic tests', () => {
```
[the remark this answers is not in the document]
digeff: I was planning on adding some tests. I agree this doesn't make any sense right now.
I'll remove it, and re-add it when I do add more tests :)

test/int/featureBasedSuits/hitCountBreakpointTests.test.ts
```diff
@@ -0,0 +1,160 @@
...
+            await setStateBreakpoint.unset();
+            await setNewValBreakpoint.unset();
+            await stepInBreakpoint.unset();
+        });
+
+        puppeteerTest("Hit count breakpoints = 3, = 4 and = 5 set in batch pause on the button's 3rd, 4th and 5th clicks", suiteContext, async (_context, page) => {
+            const incBtn = await page.waitForSelector('#incrementBtn');
+
+            const breakpoints = BreakpointsWizard.create(suiteContext.debugClient, reactTestSpecification);
+            const counterBreakpoints = breakpoints.at('Counter.jsx');
```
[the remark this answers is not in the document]
digeff: That is a good point.

I don't think we'll use the BreakpointsWizard in *every single test*. I do think we'll use BreakpointsWizard in most tests. I also think we might end up creating some other wizards or utilities for other things too.

What parameters would you pass to puppeteerSuite() so the "client" can initialize the BreakpointsWizard, or potentially another wizard/utility? Yet without forcing every single test to do it (Some loaded scripts tests won't require setting any breakpoints)?

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
digeff: Removed

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
digeff: What do you think of IBatchOrNotState?

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
digeff: Could you please elaborate on what advantages you see if we change this to use the same instance on every call?

EricCornelson: Code clarity, mainly. Something like:

`this.breakpointUpdater.send(newBreakpoints)`

is much easier to understand at a quick glance than the complicated constructor->update().

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
EricCornelson: In general I think of words like "or", "and", etc. when naming things as a code smell, as usually it's an indication that either we're doing too much with one object (violating single responsibility principle), or, as in this case, not clearly articulating what an object's responsibility _actually_ is.

`IBatchOrNotState` still tells me nothing about what the interface's responsibility actually is, only what it's possible values can be. That means that the only way I can truly understand what the interface is for is to go and look at the implementing classes.  The name of an interface should describe its purpose. What I see is that implementers of this interface do two things:

1) Define the strategy for breakpoint updates (batch vs individual)
2) Manage the current state of the breakpoints (e.g. which ones are set vs. not set)

Of those, it's the strategy that drives the behavior, and how we manage state, so I think an even better name would be something like `ISetBreakpointStrategy` or just `IBreakpointStrategy`, since we're using the strategy pattern to change the behavior of the system based on the implementation we choose.

digeff: What do you think of IBreakpointsBatchingStrategy?

EricCornelson: I like it

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
digeff: Removed!

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
digeff: Removed!

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
digeff: Changed

test/int/featureBasedSuits/hitCountBreakpointTests.test.ts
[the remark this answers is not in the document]
EricCornelson: Hmm, for now I think we can hold off as we don't have a lot of different utilities, but in the future if we're going to be adding more stuff, it sounds to me like a good place for a builder pattern, where in creation of the suite, we could add modules that would be able to add functionality to the tests within that suite.

Something like:


    suiteBuilder()
        .add(new PuppeteerThing())
        .add(new BreakpointWizard())
        .build( (context) => {
             //test code here, and utils accessible from the context?
        });

Might have to think about how to make that type-safe, but it would make things much more composeable than the current `puppeteerSuite`, and for each suite we could add only the functionality required for those tests.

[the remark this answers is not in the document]
digeff: That sounds like a great idea! Let's keep an eye of how the tests evolve, and look into this after we know if we need more utilities/wizards or not...

PR #819, opened 2019-04-22
simple tests for loadSources panel

Shenniey:


test/int/sources/sources.noBreakpoints.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { loadedSourcesContainsScript, createScriptItemFromInfo } from './sources.utils';
+
+// We are using the "Simple" project app inside the testdata directory
+const DATA_ROOT = testSetup.DATA_ROOT;
+const SIMPLE_PROJECT_ROOT = path.join(DATA_ROOT, 'simple');
+const TEST_SPEC = new TestProjectSpec( { projectRoot: SIMPLE_PROJECT_ROOT } );
+
+puppeteerSuite('Static HTML Pages Without Breakpoints', TEST_SPEC, (suiteContext) => {
+    puppeteerTest('Static webpage with single loaded script', suiteContext, async (context) => {
+        let scriptToMatch = createScriptItemFromInfo('app.js', TEST_SPEC.props.url + 'app.js');
```
digeff: What do you think of creating an utility to be able to do something like this:
		new LoadedSourcesWizard(context.debugClient).assertSourcesAre(`
			${TEST_SPEC.props.url}/app.js
			${TEST_SPEC.props.url}/otherFile.ts
			`);
(We'll need to figure out how to eventually make the paths work on Mac too)

test/int/sources/sources.noBreakpoints.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+// We are using the "Simple" project app inside the testdata directory
+const DATA_ROOT = testSetup.DATA_ROOT;
+const SIMPLE_PROJECT_ROOT = path.join(DATA_ROOT, 'simple');
+const TEST_SPEC = new TestProjectSpec( { projectRoot: SIMPLE_PROJECT_ROOT } );
+const sourceChecker = new sourcesHelper.SourcesChecker();
+
+puppeteerSuite('Static HTML Pages Without Breakpoints', TEST_SPEC, (suiteContext) => {
+    puppeteerTest('Receive loaded source events after page load', suiteContext, async (context, page) => {
+        sourceChecker.resetLoadedSourcesCount();
```
digeff: I think that creating a new instance for each tests is less error-prone than having to reset it.

test/int/sources/sources.utils.ts
```diff
@@ -0,0 +1,46 @@
+import { DebugProtocol } from 'vscode-debugprotocol';
+import { ValidatedMap } from '../core-v2/chrome/collections/validatedMap';
+import assert = require('assert');
+
+export type Script = DebugProtocol.Source;
+export type LoadedSourceEvent = DebugProtocol.LoadedSourceEvent;
+
+export async function loadedSourcesContainsScript(loadedSources: Script[], scriptToFind: Script) : Promise<boolean> {
+    let sourcesWithGivenNameAndPath = loadedSources.filter(source => (scriptToFind.name === source.name && scriptToFind.path === source.path));
```
digeff: We should test all the properties that either VS or VS Code rely on.

test/int/sources/sources.noBreakpoints.test.ts
```diff
@@ -0,0 +1,71 @@
...
+                sourceChecker.updateSourcesCount(1);
+                sourceChecker.assertLoadedSourceCountIs(1);
+            })
+        ]);
+
+        let dynamicScript2 = 'console.log(\'hello again!\')';
+
+        await Promise.all([
+            page.evaluate(dynamicScript2),
+            context.debugClient.waitForEvent('loadedSource').then(event => {
```
digeff: This code is similar to the code on line 22. Is there any way we can avoid duplicating it?

PR #828, opened 2019-04-30
Added test for different combinations of hit count breakpoints operators and numbers and formats

digeff:
- Merge methodsCalledLogger.ts updates from v2

- hitCountBreakpointTests.test.ts: Updated hit count bps test to only test the breakpoints and not the stack traces. Also added boundText so the bp can specify the actual location if it's different from the original bp location

- hitCountBreakpointsDifferentOperators.test.ts: Test different operators, numbers, and formats for hit count bps

- waitUntilReadyWithTimeout: Some of the new test can take up to 10 seconds on my machine. Updating timeout considering that the build machines are slower

- breakpointsWizard.ts: Updated logic to store a list of the events to be consumed, instead of the last one. This is needed for when we resume and immediately hit a breakpoint. We need to store both events.

- breakpointsAssertions.ts: Added option to validate some local variables value. Updated assertions to test all the properties of the breakpoints and their bound locations. Stack trace and variables verifications are optional now.

- variablesAssertions.ts: Added code to verify the contents of local variables

- testdata/react_with_loop/: New react project with a loop to test hit count breakpoints.

- testdata/react_with_loop/src/Counter.jsx: Only modification to the react project. Added the loop method.


testdata/react_with_loop/dist/src/Running.js
```diff
@@ -0,0 +1,10 @@
+
+// a script to keep running forever
+
+let num = 0;
+export function runForever() {
+    setTimeout(() => {
+        num++;
+        runForever();
```
roblourens: Could use `setInterval` too

testdata/react_with_loop/package.json
```diff
@@ -0,0 +1,123 @@
...
+      "ts",
+      "web.tsx",
+      "tsx",
+      "json",
+      "web.jsx",
+      "jsx",
+      "node"
+    ],
+    "watchPlugins": [
+      "C:\\development\\react_test_project\\my-app\\node_modules\\jest-watch-typeahead\\filename.js",
```
roblourens: Are these paths specific to your machine? It probably doesn't matter for this test scenario though? I don't know what these are for.

testdata/react_with_loop/scripts/build.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+
+// Do this as the first thing so that any code reading it knows the right env.
+process.env.BABEL_ENV = 'production';
```
roblourens: These scripts just come with the react app right? Did you use `create-react-app` or something else?

test/int/featureBasedSuits/hitCountBreakpointsDifferentOperators.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        { condition: '= 2', iterationsExpectedToPause: [2] },
+        { condition: '= 12', iterationsExpectedToPause: [12] },
+        { condition: '> 0', iterationsExpectedToPause: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '> 1', iterationsExpectedToPause: [2, 3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '>\t2', iterationsExpectedToPause: [3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '> 187', iterationsExpectedToPause: [188, 189, 190, 191] },
+        { condition: '>=   0', iterationsExpectedToPause: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '>= 1', iterationsExpectedToPause: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '>= 2', iterationsExpectedToPause: [2, 3, 4, 5, 6, 7, 8, 9, 10] },
+        { condition: '>= 37', iterationsExpectedToPause: [37, 38, 39] },
```
roblourens: How many iterations are there? I don't follow why this only goes up to 39 when others are testing up to higher numbers

testdata/react_with_loop/scripts/build.js
digeff: I copied the app we had on the /react/ folder, and just removed the stepIn method and added the loop method instead. Every other file should be the same as far as I'm aware.

roblourens: Ok, it's no problem.

testdata/react_with_loop/dist/src/Running.js
digeff: I copy these files from the existing react project. I'm inclined to leave it as-is unless we have a good reason to change them

testdata/react_with_loop/package.json
digeff: Same as above

test/int/featureBasedSuits/hitCountBreakpointsDifferentOperators.test.ts
digeff: If I added a 40 after the 39, the test would test one more iteration and work properly.

There are 200 iterations in the loop. I just test enough of them to make sure that the condition is working as expected.

I coded the test like this on purpose, because hitting 200 breakpoints for the >= 0 condition is probably not worth it...

roblourens: But looking at that, and `{ condition: '<= 1', iterationsExpectedToPause: [1] },`, I don't see how it's testing that you are _not_ pausing on `2`, but the first test isn't testing that you are not pausing on `40`, if that makes sense.

roblourens: If it pauses on `2`, will the test fail? It doesn't look like it

digeff: That's a great point. Thank you for catching it :)

digeff: Added a new noMorePausesAfterwards parameter to verify that we don't pause on the 2nd iteration for <= 1 and other similar cases.

test/int/featureBasedSuits/hitCountBreakpointsDifferentOperators.test.ts
```diff
@@ -63,23 +64,22 @@ puppeteerSuite('Hit count breakpoints combinations', reactWithLoopTestSpecificat
...
+                /**
+                 * The iterationNumber variable counts in the js-debuggee code how many times the loop was executed. We verify
+                 * the value of this variable to validate that a bp with = 12 paused on the 12th iteration rather than on the 1st one
+                 * (The breakpoint is located in the same place in both iterations, so we need to use state to differenciate between those two cases)
+                 */
+                await setStateBreakpoint.assertIsHitThenResume({ variables: { 'iterationNumber': `${nextIterationToPause}` } });
+            }
 
+            if (conditionConfiguration.noMorePausesAfterwards) {
+                await breakpoints.assertNotPaused();
```
roblourens: I'm curious how this works, is it "the app is not paused right now" or "the app doesn't pause within the next x ms" or something else?

roblourens: Oh it's in this PR. You might rename it to `noPauseWithin1s` or something to that effect?

digeff: Changed to waitAndAssertNotPaused(millisecondsToWaitForPauses = 1000 /*ms*/). What do you think?

I'd rather not put the exact time on the method, because we might need to change it for the build server...

roblourens: Yeah that sounds perfect.

roblourens:
Is this a test that can go into master? I just realized it's for v2.

PR #831, opened 2019-05-03
Add tests for launching/attaching to chrome

EricCornelson:
Adding tests for launching chrome (relies on https://github.com/Microsoft/vscode-chrome-debug-core/pull/438)

### DIFF
--- a/test/int/adapter.test.ts
+++ b/test/int/adapter.test.ts
@@ -24,7 +24,7 @@ suite('Chrome Debug Adapter etc', () => {
     let server: HttpOrHttpsServer | null;
 
     setup(function () {
-        return testSetup.setup(this, )
+        return testSetup.setup(this)
             .then(_dc => dc = _dc);
     });
 
@@ -114,7 +114,7 @@ suite('Chrome Debug Adapter etc', () => {
 
         test('Should throw error when launching if chrome debug port is in use', async () => {
             // browser already launched to the default port, and navigated away from about:blank
-            const browser = await puppeteer.launch(<puppeteer.LaunchOptions>{ headless: false, args: ['http://localhost:7890', '--remote-debugging-port=9222'] });
+            const browser = await puppeteer.launch({ headless: false, args: ['http://localhost:7890', '--remote-debugging-port=9222'] });
             try {
                 await Promise.all([
                     dc.configurationSequence(),
--- a/test/int/adapter.test.ts
+++ b/test/int/adapter.test.ts
@@ -12,10 +12,14 @@ import { ExtendedDebugClient } from 'vscode-chrome-debug-core-testsupport';
 import * as testSetup from './testSetup';
 import { HttpOrHttpsServer } from './types/server';
 import { isWindows } from './testSetup';
+import * as puppeteer from 'puppeteer';
+import { IAttachRequestArgs } from '../../../vscode-chrome-debug-core/lib/src/debugAdapterInterfaces';
+import { execSync } from 'child_process';
+import { expect } from 'chai';
 
 const DATA_ROOT = testSetup.DATA_ROOT;
 
-suite('Chrome Debug Adapter etc', function () {
+suite('Chrome Debug Adapter etc', () => {
     let dc: ExtendedDebugClient;
     let server: HttpOrHttpsServer | null;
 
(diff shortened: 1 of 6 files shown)


test/int/adapter.test.ts
```diff
@@ -85,5 +89,52 @@ suite('Chrome Debug Adapter etc', function () {
...
             ]);
         });
+
+        test('Should attach to existing instance of chrome and break on debugger statement', async () => {
+            const testProjectRoot = path.join(DATA_ROOT, 'intervalDebugger');
+            const breakFile = path.join(testProjectRoot, 'src/app.ts');
+            const DEBUGGER_LINE = 2;
+
+            server = createServer({ root: testProjectRoot });
+            server.listen(7890);
```
digeff: This code is similar to line 118 and to the should stop on debugger statement in http://localhost test
Is there any way we can avoid duplicating it?

test/int/adapter.test.ts
```diff
@@ -85,5 +89,52 @@ suite('Chrome Debug Adapter etc', function () {
...
+                ]);
+            } catch (err) {
+                expect(err.message).to.satisfy( (x: string) => x.startsWith('Cannot connect to runtime process, timeout after 2000 ms'));
+            }
+
+            await browser.close();
+
+            console.log('killing all instances of chrome');
+            // force kill chrome here, as it will be left open by the debug adapter (same behavior as v1)
+            const killCmd = (process.platform === 'win32') ? 'taskkill /F /IM chrome.exe /T' : 'killall chrome';
```
digeff: This code is similar to the code in testSetup.tearDown. Would it be possible to unify it?

Is this behavior undesirable? If so, we should add a TODO to fix it after we release v2.

test/int/adapter.test.ts
```diff
@@ -12,10 +12,14 @@ import { ExtendedDebugClient } from 'vscode-chrome-debug-core-testsupport';
...
 import { isWindows } from './testSetup';
+import * as puppeteer from 'puppeteer';
+import { IAttachRequestArgs } from '../../../vscode-chrome-debug-core/lib/src/debugAdapterInterfaces';
+import { execSync } from 'child_process';
+import { expect } from 'chai';
 
 const DATA_ROOT = testSetup.DATA_ROOT;
 
-suite('Chrome Debug Adapter etc', function () {
+suite('Chrome Debug Adapter etc', () => {
```
digeff: Mocha suggests using normal functions instead of arrow functions, so we can access the this pointer if needed: https://mochajs.org/#arrow-functions

test/int/adapter.test.ts
```diff
@@ -85,5 +89,52 @@ suite('Chrome Debug Adapter etc', function () {
...
         });
+
+        test('Should attach to existing instance of chrome and break on debugger statement', async () => {
+            const testProjectRoot = path.join(DATA_ROOT, 'intervalDebugger');
+            const breakFile = path.join(testProjectRoot, 'src/app.ts');
+            const DEBUGGER_LINE = 2;
+
+            server = createServer({ root: testProjectRoot });
+            server.listen(7890);
+            const browser = await puppeteer.launch(<puppeteer.LaunchOptions>{ headless: false, args: ['http://localhost:7890', '--remote-debugging-port=7777'] });
```
digeff: Why is this cast needed?

test/int/adapter.test.ts
roblourens: I think you can still do this with arrow functions: https://stackoverflow.com/questions/15971167/how-to-increase-timeout-for-a-single-test-case-in-mocha. I would vote for using arrow functions here.

EricCornelson: Yeah, the reason I changed this to an arrow function was mainly just because of our lint rules that will flag any regular `function` declarations, and were red-lining the entire test in the editor. The other option is just to add tslint exceptions for all the tests, but that's also kind of annoying. There's a big long discussion about arrow functions vs normal [here](https://github.com/mochajs/mocha/issues/2018). There's been some chatter about redesigns in mocha to better support arrow functions, but so far it doesn't seem to have gone anywhere.

EricCornelson: FWIW, Jasmine also recommends `function` over `()=>{}`
https://stackoverflow.com/questions/38915090/are-arrow-functions-recommended-in-jasmine-test-cases.

Personally, I'd prefer to use arrow functions almost everywhere since it makes javascript behave more like C#, Java, and others in regards to `this`, and of course I'm more familiar with those languages, but it seems like the JS community (at least right now) still relies pretty heavily on the current `this` behavior.

So I can go either way on this.

test/int/adapter.test.ts
EricCornelson: Not needed for anything except intellisense. For whatever reason, VS Code doesn't provide intellisense on anonymous objects, even when the type of the argument is known for that function unless you provide a type assertion. We can remove it if you'd prefer, I typically just toss them on there when I need to check what the possible values can be.

test/int/adapter.test.ts
digeff: Let's use the arrow function then. If we ever need to do something which the only way to do it is to get the this pointer, then we can change that particular call to use a function instead.

test/int/adapter.test.ts
roblourens: It can make troubleshooting easier. If you start Chrome, fail to attach, then kill Chrome, sometimes people might think that Chrome never actually started.

EricCornelson: I definitely think we need to look at changing the behavior for V2 after we get the initial release out, because what essentially happens here is that the debugger attaches properly but expects to find an "about:blank" page, which we launch to when we launch chrome. If a user already has an instance of chrome running with the debug flag, it won't have that about:blank page, and the debugger bails, even if it might have been able to recover.

For now I'll create a "kill all instance of chrome" function that will get rid of the duplicated code.

test/int/adapter.test.ts
EricCornelson: Addressed. We set this up in the `setup()` for this suite now

test/int/adapter.test.ts
```diff
@@ -72,18 +85,49 @@ suite('Chrome Debug Adapter etc', function () {
...
+            await browser.close();
+        });
+
+        test('Should throw error when launching if chrome debug port is in use', async () => {
+            // browser already launched to the default port, and navigated away from about:blank
+            const browser = await puppeteer.launch({ headless: false, args: ['http://localhost:7890', '--remote-debugging-port=9222'] });
+            try {
+                await Promise.all([
+                    dc.configurationSequence(),
+                    dc.launch({ url: 'http://localhost:7890', timeout: 2000, webRoot: testProjectRoot }),
```
digeff: I'd specify the port here as an explicit parameter, to make it clearer that we are going to be using the same port...

test/int/adapter.test.ts
```diff
@@ -72,18 +85,49 @@ suite('Chrome Debug Adapter etc', function () {
...
+                ]);
+            } catch (err) {
+                expect(err.message).to.satisfy( (x: string) => x.startsWith('Cannot connect to runtime process, timeout after 2000 ms'));
```
digeff: Will this test fail if dc.launch somehow succeeds?

test/int/adapter.test.ts
```diff
@@ -72,18 +85,49 @@ suite('Chrome Debug Adapter etc', function () {
...
+
+            await Promise.all([
+                dc.configurationSequence(),
+                dc.initializeRequest().then(_ => {
+                    return dc.attachRequest(<IAttachRequestArgs>{ url: 'http://localhost:7890', port: 7777, webRoot: testProjectRoot });
+                }),
+                dc.assertStoppedLocation('debugger_statement', { path: breakFile, line: DEBUGGER_LINE } )
+            ]);
+
+            await browser.close();
```
digeff: It might be a good idea to put the .close() in a finally block, so we'll close the browser even if the test fails...

test/testUtils.ts
```diff
@@ -49,6 +50,15 @@ export function registerLocMocks(): void {
     });
 }
 
+/**
+ * Kills all running instances of chrome on the test host
+ */
+export function killAllChrome() {
+    const killCmd = (process.platform === 'win32') ? 'taskkill /F /IM chrome.exe /T' : 'killall chrome';
```
digeff: Will this kill all the chrome.exe instances on my desktop every time I run the tests?

test/int/adapter.test.ts
EricCornelson: Good suggestion. I added a finally block for both tests that close with puppeteer.

test/int/adapter.test.ts
EricCornelson: Done

test/int/adapter.test.ts
EricCornelson: Yes. Right now since we're just baselining V1 behavior, we are expecting this to throw an exception. Once we decide it's an appropriate time to fix this, then we can change the test accordingly. I'll add a note in the test that explains this so that it's clear what we're expecting from the test

test/testUtils.ts
EricCornelson: Yes. Unless there's an easy way to get the PID of the instance that the debug adapter launched so we can just close that specific instance, I don't see another option other than just force killing all chrome instances here. Let me know if you know of a better way.

digeff: I tested on my PC and "wmic process where ExecutablePath='path goes here\\chrome-win\\chrome.exe' delete" can kill only the chrome.exe instances at a specific path

PR #840, opened 2019-05-17
Stacktrace test improvements

mrcrane:
- Refactor stacktrace tests to use the common breakpoints wizard code
- Update stacktrace tests to cover eval code test case
- Update stacktrace tests to use chai expect

### DIFF
--- a/test/int/stackTrace.test.ts
+++ b/test/int/stackTrace.test.ts
@@ -6,12 +6,12 @@ import { URL } from 'url';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { FrameworkTestContext, TestProjectSpec } from './framework/frameworkTestSupport';
 import { puppeteerSuite, puppeteerTest } from './puppeteer/puppeteerSuite';
-import { setBreakpoint } from './intTestSupport';
-import { THREAD_ID } from 'vscode-chrome-debug-core-testsupport';
+import { BreakpointsWizard as BreakpointsWizard } from './wizards/breakpoints/breakpointsWizard';
+import { IStackTraceVerifier } from './wizards/breakpoints/implementation/breakpointsAssertions';
 
 const DATA_ROOT = testSetup.DATA_ROOT;
 const SIMPLE_PROJECT_ROOT = path.join(DATA_ROOT, 'stackTrace');
-const TEST_SPEC = new TestProjectSpec( { projectRoot: SIMPLE_PROJECT_ROOT } );
+const TEST_SPEC = new TestProjectSpec( { projectRoot: SIMPLE_PROJECT_ROOT, projectSrc: SIMPLE_PROJECT_ROOT } );
 const TEST_URL = new URL(TEST_SPEC.props.url);
 
 interface ExpectedSource {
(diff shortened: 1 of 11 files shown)


testdata/stackTrace/app.js
```diff
@@ -1,9 +1,13 @@
...
 }
 
-function inner() {
+function timeoutCallback() {
+    eval("evalCallback();");
+}
+
+function evalCallback() {
     (function() {
-        console.log('Inside anonymous function'); // bpLabel: stackTraceBreakpoint
```
roblourens: I still see that bpLabel referenced elsewhere, is that a problem?

test/int/stackTrace.test.ts
```diff
@@ -138,8 +154,10 @@ puppeteerSuite.only('Stack Traces', TEST_SPEC, (suiteContext) => {
             buttonIdToClick: '#button',
             format: undefined,
             expectedFranes: [
```
digeff: fra **n** es

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
```diff
@@ -57,17 +94,28 @@ export class BreakpointsAssertions {
     public async assertIsHitThenResume(breakpoint: BreakpointWizard, verifications: IVerifications): Promise<void> {
         await this._breakpointsWizard.waitUntilPaused(breakpoint);
 
-        const stackTrace = await this.stackTrace();
-        const topFrame = stackTrace[0];
+        let stackFrameFormat: DebugProtocol.StackFrameFormat = this._defaultStackFrameFormat;
+        let stackTraceVerifier: IStackTraceVerifier | undefined = undefined;
+
+        if (verifications.stackTrace && verifications.stackTraceVerifier) {
```
digeff: What do you think, instead of passing the verifications.stackTraceVerifier, checking whether verifications.stackTrace is a string or an object, and then automatically using the corresponding verifier?

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
```diff
@@ -29,11 +34,43 @@ interface IObjectWithLocation {
...
+                expect(stackTraceResponse.body.totalFrames, `The number of stackFrames was different than the value supplied on the totalFrames field`)
+                    .to.equal(stackTraceResponse.body.stackFrames.length);
+                stackTraceResponse.body.stackFrames.forEach(frame => {
+                    // Warning: We don't currently validate frame.source.path
+                    expect(frame.source).not.to.equal(undefined);
+                    const expectedSourceNameAndLine = ` [${frame.source!.name}] Line ${frame.line}`;
+                    expect(frame.name, 'Expected the formatted name to match the source name and line supplied as individual attributes').to.endsWith(expectedSourceNameAndLine);
+                });
+
+                const stackTrace = stackTraceResponse.body.stackFrames;
```
digeff: If you move this up to line 54, you can use it for line 56 and 57

test/int/stackTrace.test.ts
```diff
@@ -55,37 +58,50 @@ function assertSourceMatches(actual: DebugProtocol.Source | undefined, expected:
...
-    assert.equal(actual.path, expectedPath, `Source path for frame ${index} does not match`);
+    expect(actual.name).to.equal(expectedName, `Frame ${index} source name`);
+    expect(actual.path).to.equal(expectedPath, `Frame ${index} source path`);
 }
 
 function assertFrameMatches(actual: DebugProtocol.StackFrame, expected: ExpectedFrame, index: number) {
-    assert.equal(actual.name, expected.name, `Name for frame ${index} does not match`);
-    assert.equal(actual.line, expected.line, `Line number for frame ${index} does not match`);
-    assert.equal(actual.column, expected.column, `Column number for frame ${index} does not match`);
+    if (expected.name) {
```
digeff: What do you think of using name for both string and regexp, and using typeof or instance of to figure out which assertion to use?

testdata/stackTrace/app.js
mrcrane: The framework tests (test/int/framework/*) still use the bpLabel approach and currently still need the labels. That is unrelated to these tests after this change.

roblourens: Oh ok

test/int/stackTrace.test.ts
```diff
@@ -2,25 +2,28 @@ import * as assert from 'assert';
...
 import * as puppeteer from 'puppeteer';
 import * as testSetup from './testSetup';
+import { expect } from 'chai';
 import { URL } from 'url';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { FrameworkTestContext, TestProjectSpec } from './framework/frameworkTestSupport';
 import { puppeteerSuite, puppeteerTest } from './puppeteer/puppeteerSuite';
-import { setBreakpoint } from './intTestSupport';
-import { THREAD_ID } from 'vscode-chrome-debug-core-testsupport';
+import { BreakpointsWizard as BreakpointsWizard } from './wizards/breakpoints/breakpointsWizard';
```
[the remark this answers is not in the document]
mrcrane: Fixed

test/int/wizards/breakpoints/implementation/internalFileBreakpointsWizard.ts
```diff
@@ -48,6 +48,12 @@ export class InternalFileBreakpointsWizard {
 
     public constructor(private readonly _breakpointsWizard: BreakpointsWizard, public readonly client: ExtendedDebugClient, public readonly filePath: string) { }
 
+    public async breakpoint(options: { text: string, name: string, boundText?: string }): Promise<BreakpointWizard> {
+        const position = await findPositionOfTextInFile(this.filePath, options.text);
```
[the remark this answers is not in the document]
mrcrane: Good idea, I refactored this

test/int/stackTrace.test.ts
mrcrane: Fixed

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
mrcrane: Fixed

test/int/stackTrace.test.ts
mrcrane: I think this is a good idea in this case. I fixed this.

test/int/wizards/breakpoints/implementation/breakpointsAssertions.ts
mrcrane: I wish we were able to check the interfaces directly, instead of checking string vs object. Otherwise I think it's a good idea. I made the change, which I think is cleaner now.

PR #845, opened 2019-05-23
Add changes from core that we need for the variables test

digeff:
Add changes from core that we need for the variables test

### DIFF
--- a/test/int/core-v2/chrome/collections/utilities.ts
+++ b/test/int/core-v2/chrome/collections/utilities.ts
@@ -1,0 +1,29 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { ValidatedMultiMap } from './validatedMultiMap';
+
+export function groupByKey<T, K>(elements: T[], obtainKey: (element: T) => K): ValidatedMultiMap<K, T> {
+    const groupped = ValidatedMultiMap.empty<K, T>();
+    elements.forEach(element => groupped.add(obtainKey(element), element));
+    return groupped;
+}
+
+export function determineOrderingOfStrings(left: string, right: string): number {
+    if (left < right) {
+        return -1;
+    } else if (left > right) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


test/int/core-v2/chrome/collections/utilities.ts
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { ValidatedMultiMap } from './validatedMultiMap';
+
+export function groupByKey<T, K>(elements: T[], obtainKey: (element: T) => K): ValidatedMultiMap<K, T> {
```
roblourens: Can these be imported from core rather than copy/pasting them?

digeff: For v2 it's really easy to import them from core.
For v1 it might be possible if we import 2 different versions of -core (both master and v2). I personally think this would not be worth the trouble.

test/int/core-v2/chrome/collections/utilities.ts
```diff
@@ -0,0 +1,29 @@
...
+    const groupped = ValidatedMultiMap.empty<K, T>();
```
EricCornelson: Spelling nit: "grouped"

test/int/core-v2/chrome/collections/utilities.ts
EricCornelson: Does it make sense to import for v2 and make the copy only for v1?

digeff: I'm open to suggestions as to how to do it. Ideally we'd be able to keep the same exact test code for both v1 and v2 to avoid having to do extra work.

test/int/core-v2/chrome/collections/utilities.ts
digeff: Fixed

test/int/core-v2/chrome/collections/utilities.ts
roblourens: I see - I'm ok with doing it this way as long as the copied code is clearly marked as coming from v2, like you've already done with the core-v2 folder, so we can clean it up later.

PR #855, opened 2019-06-06
Add more variables tests for other scopes

digeff:
Add more variables tests for other scopes
We test the same basic set of variable types in each of the possible scopes

```
Variables scopes
    √ local
    √ globals
    √ script
    √ block
    √ catch
    √ closure
    √ eval
    √ with
    √ module

```


EricCornelson:
Looks good. The only question I have is whether the order of the variables can ever change, (or if it did change, is that a bug or not).

digeff:
The order at least in some occasions matter. The this variable should always appear as the first variable for example...

PR #864, opened 2019-06-26
Merging in test updates from master

EricCornelson:


test/int/adapter.test.ts
```diff
@@ -121,6 +121,42 @@ suite('Chrome Debug Adapter etc', () => {
...
+
+            await dc.initializeRequest({
+                adapterID: 'chrome',
+                clientID: 'visualstudio',
+                linesStartAt1: true,
+                columnsStartAt1: true,
+                pathFormat: 'path'
+            });
+
+            await dc.launchRequest( { url: 'http://localhost:7890', webRoot: testProjectRoot.toLowerCase(), runtimeExecutable: puppeteer.executablePath() } as any);
```
digeff: Why do we need to do the "as any"?

EricCornelson: It's back to the "url problem". The base launch request args doesn't have a url property, but the chrome debug launch args do. The ExtendedDebugClient's `launch` method accepts `any` but because this test was calling `launchRequest` (which takes an explicit `DebugProtocol.LaunchRequestArguments`) instead of `launch`, we can't actually pass our chrome launch args (where url is a totally valid property) unless we force it to accept any, or force it with a type hint.

PR #868, opened 2019-07-01
Add test for unhandled promises

digeff:
Test that we pause on unhandled promise rejections and that this bug stays fixed: https://github.com/microsoft/vscode-chrome-debug-core/pull/466

### DIFF
--- a/test/int/wizards/pausedWizard.ts
+++ b/test/int/wizards/pausedWizard.ts
@@ -67,25 +67,6 @@ export class PausedWizard {
         });
     }
 
-    /** Wait and block until the debuggee is paused on an unhandled promise */
-    public async waitUntilPausedOnPromiseRejection(exceptionMessage: string): Promise<void> {
-        return this.waitAndConsumePausedEvent(async pauseInfo => {
-            expect(pauseInfo.description).to.equal('Paused on promise rejection');
-            expect(pauseInfo.reason).to.equal('exception');
-
-            const exceptionInfo = await this._client.exceptionInfoRequest({ threadId: THREAD_ID });
-            expect(exceptionInfo.success).to.equal(true);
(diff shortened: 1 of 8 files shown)


test/int/wizards/pausedWizard.ts
```diff
@@ -67,12 +67,31 @@ export class PausedWizard {
         });
     }
 
+    /** Wait and block until the debuggee is paused on an unhandled promise */
```
EricCornelson: Does this really belong in pausedWizard? This seems extremely specific to this test.

digeff: We are going to use a similar code when we test pausing on an exception. I think it does make sense to have it in a utility or shared place. Maybe the right thing to do would be to create an ExceptionsWizard and move it there. What do you think?

EricCornelson: I just worry about having more and more test code tightly coupled to the various "wizards", to the point that it's getting really, really hard to write a simple test that doesn't depend on a whole bunch of other stuff, and carry with it a whole bunch of assumptions that may not be relevant to that particular test. 

It also makes things much more difficult to port back to V1, where some of these assumptions don't necessarily apply.

I'm not opposed to making code shareable, but we have to do it in a way that doesn't tightly couple everything together.

Also, utility methods (in my opinion) shouldn't be doing any asserts (or expects). That should be the responsibility of each test to verify the thing it's trying to verify (and not 100 other things that are forced on it via utility code). It doesn't add any value to check the same thing in every single test.

digeff: Just to make sure I understand, your suggestion is to move this method into 
pauseOnPromisesRejections.test.ts?

PR #873, opened 2019-07-03
Add loaded sources test

digeff:
Added loaded sources test to verify script documents is shown in VS, and that we can request the source of javascript files, and .html files.

Validates bug fixed by: https://github.com/microsoft/vscode-chrome-debug-core/pull/465


roblourens:
Has a conflict

PR #874, opened 2019-07-03
Add attach test

digeff:
Added test that verifies we can attach without specifying an url parameter (similar to what VS does).
We also verify that we don't see any unhandled exceptions shortly after attaching.

Apart from the test, I modified LaunchProject, launchTestAdapter, LaunchPuppeteer to support the attach scenario, and to be able to register listeners before we issue the attach command (needed because the unhandled exceptions happen very early after attaching).

Validates bugs fixed by: https://github.com/microsoft/vscode-chrome-debug-core/pull/467


roblourens:
Conflicts and some test failure?

digeff:
The test failure is a good thing, given that the fix hasn't been merged yet: microsoft/vscode-chrome-debug-core#467 :)

PR #875, opened 2019-07-03
Now we can selectively disable assertions for the wizards

digeff:
Now we can selectively disable assertions for the wizards.

We'll experiment to see if this makes porting the tests to v1 easier.

### DIFF
--- a/test/int/utils/customChai.ts
+++ b/test/int/utils/customChai.ts
@@ -7,6 +7,7 @@
  */
 
 import * as chai from 'chai';
+import { isThisV1 } from '../testSetup';
 
 class NoopProxyHandler {
     get<T, K extends keyof T>(_target: T, _propertyKey: K, _receiver: any): any {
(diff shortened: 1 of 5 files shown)


test/int/utils/customChai.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+function defaultShouldBeNoop(_target: any, _message?: string): boolean {
+    return true;
+}
+
+export let shouldBeNoop = defaultShouldBeNoop;
+
+export const expect = customExpect;
```
roblourens: I don't understand how this is supposed to work?

digeff: expect(something) will return the noop proxy. After that, anything else you do will still return the noop proxy
expect(something) == expect(something).to == expect(something).to.be == nooproxy
So then, when you call equals(4), it's a noop, and it's not actually verifying it...

roblourens: So the goal of this is to allow disabling all assertions? From the description I thought you want to disable just certain assertions, or something?

digeff:
At the moment the tests are passing on v1, so we don't need this right now. If we need this in the future we'll reopen the PR.

PR #881, opened 2019-07-08
Merge new integration tests from V2

EricCornelson:
It's been a while since we've merge the new integration tests from V2 back into master, (sorry for the huge PR). This brings master up to date with all the latest test additions from V2, and has all tests passing.

Note: re-made the PR to fix the history and make it easier to see just the changes that were made on top of the V2 test code.


test/int/breakOnLoad.test.ts
```diff
@@ -100,7 +100,11 @@ function runCommonTests(breakOnLoadStrategy: BreakOnLoadStrategy) {
 
             if (breakOnLoadStrategy === 'instrument') {
                 await launchWithUrlAndSetBreakpoints(url, testProjectRoot, scriptPath, bpLine, bpCol);
-                await dc.assertStoppedLocation('breakpoint', { path: scriptPath, line: bpLine, column: bpCol });
+                if(testSetup.isThisV1) { // V1 says debugger_statement
+                    await dc.assertStoppedLocation('debugger_statement', { path: scriptPath, line: bpLine, column: bpCol });
```
digeff: Would it be better to do?
const expectedReason = testSetup.isThisV1 ? 'debugger_statement' : 'breakpoint';
await dc.assertStoppedLocation(expectedReason, { path: scriptPath, line: bpLine, column: bpCol });

test/int/stackTrace.test.ts
```diff
@@ -48,7 +48,9 @@ puppeteerSuite('Stack Traces', TEST_SPEC, (suiteContext) => {
             expectedFrames: [
                 { name: '(anonymous function)', line: 11, column: 9, source: { fileRelativePath: 'app.js' }, presentationHint: 'normal'},
                 { name: 'evalCallback', line: 12, column: 7, source: { fileRelativePath: 'app.js' }, presentationHint: 'normal'},
-                { name: '(eval code)', line: 1, column: 1, source: { evalCode: true }, presentationHint: 'normal'},
+                ((testSetup.isThisV2) ? // V1 produces "(anonymous function)" here
+                    { name: '(eval code)', line: 1, column: 1, source: { evalCode: true }, presentationHint: 'normal'} :
```
digeff: Would it be better to extract 'eval code' vs 'anonymous function' to a variable at the top, and use that everywhere?

EricCornelson: Well, the challenge with that here is that it's not just 'eval' code vs 'anonymous function' here, but a series of strings/regexes all of which are slightly different, e.g.:

Line 51: (eval code) vs (anonymous function)
Line 75: "/\(eval code\) \[.*VM.*]/" vs "/\(anonymous function\) \[.*VM.*]/"
Line 98: "/\(eval code\) \[.*VM.*]/" vs "(anonymous function) Line 1"
Line 126: "/\(eval code\) \[.*VM.*]/" vs "/\(anonymous function\) \[.*VM.*] Line 1/"

So we'd need to create 4 constants for each of the sites. Easier to see the differences in-line IMO

test/int/breakOnLoad.test.ts
EricCornelson: Done

test/int/stackTrace.test.ts
EricCornelson: Resolved

PR #882, opened 2019-07-09
Add test for support of file url in sources

digeff:
Test that we can hit a breakpoint in a typescript file that has a file:/// url in the source field of the source-maps.

Validates the bug fixed by: https://github.com/microsoft/vscode-chrome-debug-core/pull/473


test/int/featureBasedSuits/hitBreakpointWithUnusualSourceMap.test.ts
```diff
@@ -0,0 +1,36 @@
...
+suite('Unusual source-maps', () => {
+    testUsing(`file:/// url in sources' field`,
+        async context => {
+            const testSpec = TestProjectSpec.fromTestPath('featuresTests/unusualSourceMaps/fileUrlInSources');
+
+            // Update source-map to have a file:/// url in the sources field
+            const sourceMapPath = testSpec.src('../app.js.map');
+            const sourceMapContents = await utils.readFileP(sourceMapPath);
+            const sourceMapJSON = JSON.parse(sourceMapContents);
+            sourceMapJSON['sources'] = [`file:///${testSpec.src('../app.ts').replace(/\\/g, '/')}`];
```
EricCornelson: Why do we need to modify the source map here? It looks like `app.js.map` already uses a file:/// url

digeff: We don't know what the absolute path of app.ts is going to be in the CI machine, so we can't hardcode it. We need to update it dynamically.

EricCornelson: gotcha. That makes sense

PR #883, opened 2019-07-09
Validate that we support launching urls with escaping that is not needed

digeff:
New test that validates that we can hit breakpoints if the url has an escaped character like file:///c%3A/myproject/index.html

Validates the bug fixed by: https://github.com/microsoft/vscode-chrome-debug-core/pull/474


testdata/featuresTests/unusualLaunchJson/urlUsesEscapedCharacter/app.ts
```diff
@@ -0,0 +1,13 @@
+function runCode() {
```
EricCornelson: Is it necessary to create a new test project just for this test? Could we re-use one of the existing ones?

digeff: What disadvantage do you see in this test having it's own test project?

EricCornelson: Just excess bloat of the testdata/ dir, and the repo in general. This test really only cares about the launch.json params, and whether or not we can hit that url, so the test project itself is superfluous. We already have a bunch of little test projects that would meet the same need.

Why add more stuff to maintain when we don't need to?

digeff: I'm okay with trying to reuse projects. The existing ones have very specific behaviors that we want for the individual tests, so I don't see a great candidate to reuse it. I can move this project to a reusableProjects\ folder so people know this project doesn't have any particular behaviors, and they can reuse it for future tests. What do you think?

digeff: Changed to use simple project as discussed offline

PR #885, opened 2019-07-09
Add test to verify that we show the paused overlay when pausing

digeff:
Validate we show the paused overlay when we pause on a debugger; statement, and we remove it afterwards

Validates bug fixed by https://github.com/microsoft/vscode-chrome-debug-core/pull/476


test/int/featureBasedSuits/pauseOverlayIsShown.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    await utils.promiseTimeout(undefined, 500);
+
+    const logFilePath = launchArgs().logFilePath!;
+    const logFileContents = await utils.readFileP(logFilePath);
+    const lines = logFileContents.split('\n');
+    const lastEvent = _.findLast(lines, line => line.indexOf('Overlay.setPausedInDebuggerMessage') >= 0);
+    expect(lastEvent).to.not.equal(undefined);
+
+    // We are trying to match this string: Overlay.setPausedInDebuggerMessage\",\"params\":{ <contents here> }
+    const matches = lastEvent!.match(/Overlay\.setPausedInDebuggerMessage\\",\\"params\\":\{([^}]*)\}/);
```
EricCornelson: This doesn't actually test that we sent the correct message to the target, only that we logged something. 

If we truly want to test that we're sending the correct message to the target, we either need to intercept the messages in a "sniffer" or a proxy over the chrome api classes, or use a "fake target" which can validate that we sent the right message.

digeff: The logger is basically sniffing the messages, and logging them to a file... It's true that we can have a bug on the logger, and log something we didn't send. We could also have a bug in our sniffer or proxy, and report that we sent something we didn't sent...

EricCornelson: > The logger is basically sniffing the messages, and logging them to a file

Fair point, but my point isn't about having bugs that cause the test to fail, but rather avoiding having our tests coupled to our logging implementation. It makes the test brittle, because there are a whole host of scenarios where a small change to the logging implementation could break the test.

You've mentioned before how important it is to make sure our tests are deterministic, and this one isn't. What happens if the test instances are extremely slow, and our 500ms timeout isn't enough?

digeff: If that happens, we can change it to poll several times for up to 10 seconds or so...

test/int/featureBasedSuits/pauseOverlayIsShown.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Wait a little to give the log file time to get written...
+    // Warning: If this test starts failing because 500 ms being too little time, we should change the logic to read the file, and retry a few times to see if the assertion passes eventually
+    // If that doesn't work either, we'll need to do something less hacky like implementing a sniffer or proxy of the protocol, and get the information directly from there instead of
+    // reading it from a file
+    await utils.promiseTimeout(undefined, 500);
+
+    const logFilePath = launchArgs().logFilePath!;
+    const logFileContents = await utils.readFileP(logFilePath);
+    const lines = logFileContents.split('\n');
+    const lastEvent = _.findLast(lines, line => line.indexOf('Overlay.setPausedInDebuggerMessage') >= 0);
```
roblourens: I don't understand why you have to get it this way, isn't the test framework receiving all debug protocol traffic from the DA?

roblourens: Never mind, this is a message to chrome

PR #888, opened 2019-07-15
Updating CI job to run integration tests in V1

EricCornelson:


.vsts/job.yml
```diff
@@ -61,14 +61,14 @@ jobs:
...
-    - task: Npm@1
-      condition: false
-      displayName: npm frameworkTest
-      inputs:
-        command: custom
-        verbose: false
-        customCommand: 'run frameworkTest'
+  #- ${{ if eq(parameters.name, 'Windows') }}:
+  - task: Npm@1
+    condition: false
```
digeff: Shouldn't be remove this to have the tests run?

EricCornelson: Yep! :)

.vsts/job.yml
```diff
@@ -102,10 +95,25 @@ jobs:
...
     - task: PublishTestResults@2
+      condition: always()
       inputs:
         testResultsFormat: 'JUnit' # Options: JUnit, NUnit, VSTest, xUnit, cTest
         testResultsFiles: '**/test-*.xml'
+        publishRunAttachments: true
 
+  - ${{ if eq(parameters.name, 'Windows') }}:
+    - task: PublishPipelineArtifact@0
+      condition: always()
```
roblourens: You don't need condition. You could use that instead of the if block for windows if you want

EricCornelson: Done

PR #896, opened 2019-07-25
Fix dependencies

digeff:
Fix dependencies


roblourens:
Can one of you look at why the tests are flaky?

digeff:
28 failures O_o... That's a lot.

I'll look into it.

PR #901, opened 2019-07-29
Fix CI tests

digeff:
We were hitting a known issue of http-server: 0.10.0: https://github.com/http-party/http-server/issues/525
Reverting to 0.9.0 for the time being...

Also added some logging while diagnosing the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "gulp-tslint": "^8.1.3",
     "gulp-typescript": "^4.0.1",
     "gulp-util": "^3.0.8",
-    "http-server": "^0.9.0",
+    "http-server": "0.9.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "gulp-tslint": "^8.1.3",
     "gulp-typescript": "^4.0.1",
     "gulp-util": "^3.0.8",
-    "http-server": "^0.10.0",
+    "http-server": "^0.9.0",
```
EricCornelson: Pin the version?

digeff: Done

thornjad:
That issue in http-server is fixed in version 0.12.0 if you'd like to upgrade

PR #908, opened 2019-08-08
Adding support (and tests) for launching with port 0

EricCornelson:
Depends on https://github.com/microsoft/vscode-chrome-debug-core/pull/508 and adds tests for that use case.

### DIFF
--- a/src/chromeDebugAdapter.ts
+++ b/src/chromeDebugAdapter.ts
@@ -19,6 +19,7 @@ import * as utils from './utils';
 import * as errors from './errors';
 
 import { FinishedStartingUpEventArguments } from 'vscode-chrome-debug-core/lib/src/executionTimingsReporter';
+import { ChromeProvidedPortConnection } from './chromeProvidedPortConnection';
 
 // Keep in sync with sourceMapPathOverrides package.json default
 const DefaultWebSourceMapPathOverrides: ISourceMapPathOverrides = {
@@ -41,6 +42,7 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
     private _chromePID: number;
     private _userRequestedUrl: string;
     private _doesHostSupportLaunchUnelevatedProcessRequest: boolean;
+    protected _chromeConnection: ChromeProvidedPortConnection;
 
     public initialize(args: IExtendedInitializeRequestArguments): VSDebugProtocolCapabilities {
         this._overlayHelper = new utils.DebounceHelper(/*timeoutMs=*/200);
@@ -86,7 +88,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
             }
 
             // Start with remote debugging enabled
-            const port = args.port || 9222;
+            // allow port = 0
+            let port = (args.port !== undefined) ? args.port : 9222;
             const chromeArgs: string[] = [];
             const chromeEnv: coreUtils.IStringDictionary<string> = args.env || null;
             const chromeWorkingDir: string = args.cwd || null;
@@ -113,6 +116,7 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
 
             if (args.userDataDir) {
                 chromeArgs.push('--user-data-dir=' + args.userDataDir);
+                this._chromeConnection.setUserDataDir(args.userDataDir);
             }
 
             if (args._clientOverlayPausedMessage) {
--- a/src/chromeProvidedPortConnection.ts
+++ b/src/chromeProvidedPortConnection.ts
@@ -1,0 +1,58 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { chromeConnection } from 'vscode-chrome-debug-core';
+import { utils, chromeUtils } from 'vscode-chrome-debug-core';
+import { logger } from 'vscode-chrome-debug-core';
+import * as errors from './errors';
+
+import * as nls from 'vscode-nls';
+const localize = nls.loadMessageBundle();
+
+/**
+ * Chrome connection class that supports launching with --remote-debugging-port=0 to get a random port for the debug session
+ */
+export class ChromeProvidedPortConnection extends chromeConnection.ChromeConnection {
+
+    private userDataDir = undefined;
+    setUserDataDir(userDataDir: string) {
+        this.userDataDir = userDataDir;
+    }
+
+    /**
+     * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
+     * If we launched with port = 0, then this method will read the launched port from the user data directory, and wait until the port is open
+     * before calling super.attach
+     */
+    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout = chromeConnection.ChromeConnection.ATTACH_TIMEOUT, extraCRDPChannelPort?: number): Promise<void> {
+        if (port === 0 && (this.userDataDir === undefined || this.userDataDir === '')) return errors.chromeProvidedPortWithoutUserDataDir();
+        return utils.retryAsync(async () => {
+            const launchedPort = (port === 0 && this.userDataDir) ? await this.getLaunchedPort(address, this.userDataDir) : port;
+            return launchedPort;
+        }, timeout, /*intervalDelay=*/200)
+            .catch(err =>  Promise.reject(err))
+            .then(launchedPort => {
+                return super.attach(address, launchedPort, targetUrl, timeout, extraCRDPChannelPort);
+        });
+    }
+
+    /**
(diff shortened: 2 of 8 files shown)


src/chromeDebugAdapter.ts
```diff
@@ -19,6 +19,7 @@ import * as utils from './utils';
 import * as errors from './errors';
 
 import { FinishedStartingUpEventArguments } from 'vscode-chrome-debug-core/lib/src/executionTimingsReporter';
+import { isUndefined } from 'util';
```
digeff: This is deprecated: https://nodejs.org/api/util.html#util_util_isundefined_object

src/chromeDebugAdapter.ts
```diff
@@ -147,8 +149,9 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
                 });
             }
 
+            const userDataDir = (args.userDataDir || undefined);
```
digeff: In which scenario do we use this undefined?

test/int/adapter.test.ts
```diff
@@ -185,5 +185,38 @@ suite('Chrome Debug Adapter etc', () => {
...
+
+        test('Should launch successfully on port 0', async () => {
+            // browser already launched to the default port, and navigated away from about:blank
+            const remoteDebuggingPort = 0;
+            await Promise.all([
+                dc.configurationSequence(),
+                dc.launch({ url: 'http://localhost:7890', timeout: 2000, webRoot: testProjectRoot, port: remoteDebuggingPort }),
+            ]);
+            // send some commands over CDP to verify the connection
+            const pauseResult = await dc.pauseRequest({threadId: 0});
```
digeff: I think that an evaluate window.location.href is probably a more reliable way to test that you are indeed connected to Chrome.

test/int/adapter.test.ts
```diff
@@ -185,5 +185,38 @@ suite('Chrome Debug Adapter etc', () => {
...
+        });
+
+        test('Should launch successfully on port 0, even when a browser instance is already running', async () => {
+            // browser already launched to the default port, and navigated away from about:blank
+            const remoteDebuggingPort = 0;
+            const dataDir = path.join(__dirname, 'testDataDir');
+            const browser = await puppeteer.launch({ headless: false, args: ['http://localhost:7890', `--user-data-dir=${dataDir}`, `--remote-debugging-port=${remoteDebuggingPort}`] });
+            await Promise.all([
+                dc.configurationSequence(),
+                dc.launch({ url: 'http://localhost:7890', timeout: 2000, webRoot: testProjectRoot, port: remoteDebuggingPort, userDataDir: dataDir }),
```
digeff: I'd launch a different url/page to verify that we connect to the right tab

EricCornelson: Good point, done

src/chromeDebugAdapter.ts
EricCornelson: Fixed

src/chromeDebugAdapter.ts
EricCornelson: There is a case where it can still be undefined (i.e. when userDataDir is not passed, and runtimeExecutable is not set). At this point, we know that it will be either a string or undefined (not a boolean).

test/int/adapter.test.ts
EricCornelson: Not sure I understand this one. How would I check window.location.href, and why would that be more reliable?

EricCornelson: Talked offline. This is done

src/chromeDebugAdapter.ts
digeff: I'm confused as to how we use undefined then. If args.userDataDir is undefined then const userDataDir = args.userDataDir; will have the same effect as this line.

src/chromeDebugAdapter.ts
digeff: I don't see this change...

src/chromeDebugAdapter.ts
```diff
@@ -86,7 +87,8 @@ export class ChromeDebugAdapter extends CoreDebugAdapter {
             }
 
             // Start with remote debugging enabled
-            const port = args.port || 9222;
+            // allow port = 0
+            let port = !(args.port === undefined) ? args.port : 9222;
```
digeff: Any reason not to use !== ?

test/int/adapter.test.ts
```diff
@@ -185,5 +185,42 @@ suite('Chrome Debug Adapter etc', () => {
...
+            ]);
+            // send some commands over CDP to verify the connection
+            const response = await dc.evaluateRequest({
+                context: 'repl',
+                expression: 'window.location.href'
+            });
+
+            expect(response.body.result).to.equal('"http://localhost:7890/"');
+            await browser.close();
```
digeff: Should this be inside a finally?

src/chromeDebugAdapter.ts
EricCornelson: Changed with refactoring so this shouldn't be an issue anymore

src/chromeDebugAdapter.ts
EricCornelson: Nope! Changed

test/int/adapter.test.ts
EricCornelson: Good point. Done

src/errors.ts
```diff
@@ -18,3 +18,20 @@ export function getNotExistErrorResponse(attribute: string, path: string): Promi
...
+        id: 2008,
+        format: localize('random.port.no.userdatadir', 'When the remote debugging port is set to 0, you must also provide the "userDataDir" launch argument'),
+        sendTelemetry: true
+    }));
+}
+
+export function couldNotConnectToPort(address: string, port: number) {
+    return Promise.reject(new ErrorWithMessage(<DebugProtocol.Message>{
+        id: 2008,
+        format: localize('launch.port.not.open', 'Could not open a connection to Chrome DevTools at: {address}:{port}', '{address}', '{port}'),
```
digeff: Would it be clearer for the user if we say we are connecting to just "Chrome"?

src/randomPortChromeConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+import { logger } from "vscode-chrome-debug-core";
+import * as errors from './errors';
+
+import * as nls from 'vscode-nls';
+const localize = nls.loadMessageBundle();
+
+/**
+ * Chrome connection class that supports launching with --remote-debugging-port=0 to get a random port for the debug session
+ */
+export class RandomPortChromeConnection extends chromeConnection.ChromeConnection {
```
digeff: I'd rename this to ConnectionWithChromeProvidedPort or ConnectionWithChromeChoosenPort

src/randomPortChromeConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
+     * If we launched with port = 0, then this method will read the launched port from the user data directory, and wait until the port is open
+     * before calling super.attach
+     */
+    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout = chromeConnection.ChromeConnection.ATTACH_TIMEOUT, extraCRDPChannelPort?: number): Promise<void> {
+        if(port === 0 && (this.userDataDir === undefined)) return errors.randomPortWithoutUserDataDir();
+        return utils.retryAsync(async () => {
+            const launchedPort = (port === 0 && this.userDataDir) ? await this.getLaunchedPort(address, this.userDataDir) : port;
```
digeff: Will this work as espected for this.userDataDir === ""?

src/errors.ts
EricCornelson: done

src/randomPortChromeConnection.ts
EricCornelson: done

src/randomPortChromeConnection.ts
EricCornelson: Good point. Added a check for non-empty string.

src/chromeProvidedPortConnection.ts
```diff
@@ -0,0 +1,53 @@
+import { chromeConnection } from 'vscode-chrome-debug-core';
```
digeff: Missing Copyright

EricCornelson: yikes! fixed

test/int/adapter.test.ts
```diff
@@ -185,5 +187,53 @@ suite('Chrome Debug Adapter etc', () => {
...
+                await browser.close();
+            }
+        });
     });
-});
\ No newline at end of file
+});
+
+
+async function waitForUrl(dc: DebugClient, url: string): Promise<DebugProtocol.EvaluateResponse> {
```
roblourens: Would you rather wait for the page navigated event or something?

roblourens: It doesn't really matter for a test...

EricCornelson:
Yep, I think we're good to go

PR #914, opened 2019-08-20
Multiple breakpoints test and fix existing test

digeff:
* hitCountBreakpointsDifferentOperators.test.ts:
1. Removed some invalid conditions that are actually valid. The test passed before because we were doing a try {} catch { assert() } so we didn't throw an exception and the test passed for the valid conditions
2. Originally this test was testing that setBreakpoint fails when a condition is invalid. Changed it to expect setBreakpoints to succeed and return a non verified

* multipleBreakpoints.test.ts: New test verifying that if we set 3 breakpoints and 1 is invalid, the other 2 are set and hit correctly

* breakpointWizard.ts: Added the assertIsNotVerified method. Also removed the state pattern to simplify the code as part of this change.

* breakpointsWizard.ts: Moved the methods assertIsHitThenResumeWhen and assertIsHitThenResume from breakpointsAssertions.ts to add support for multiple breakpoints

* fileBreakpointsWizard.ts: Added method to set an unverified hit count breakpoint

* batchingUpdatesState.ts: Added support for assertIsNotVerified

* breakpointsAssertions.ts: 
1. Moved assertIsHitThenResumeWhen and assertIsHitThenResume to breakpointsWizard.ts.
2. Changed assertLocationMatchesExpected to a exported function assertMatchesBreakpointLocation so we can call it from breakpointsWizard.ts
3. Added assertIsNotVerified

### DIFF
--- a/test/int/featureBasedSuits/multipleBreakpoints.test.ts
+++ b/test/int/featureBasedSuits/multipleBreakpoints.test.ts
@@ -1,0 +1,40 @@
+import { puppeteerSuite, puppeteerTest } from '../puppeteer/puppeteerSuite';
+
+import { reactTestSpecification } from '../resources/resourceProjects';
+import { BreakpointsWizard } from '../wizards/breakpoints/breakpointsWizard';
+
+puppeteerSuite('Multiple breakpoints on a React project', reactTestSpecification, (suiteContext) => {
+    puppeteerTest('Can hit two valid breakpoints, while we set them with an invalid hit count breakpoints', suiteContext, async (_context, page) => {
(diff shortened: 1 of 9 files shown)


test/int/featureBasedSuits/multipleBreakpoints.test.ts
```diff
@@ -0,0 +1,40 @@
+import { puppeteerSuite, puppeteerTest } from '../puppeteer/puppeteerSuite';
```
roblourens: copyright :)

digeff: Fixed

PR #920, opened 2019-08-29
Merging chrome-provided port feature from v1

EricCornelson:
Depends on https://github.com/microsoft/vscode-chrome-debug-core/pull/520


test/int/adapter.test.ts
```diff
@@ -15,8 +15,9 @@ import { isWindows } from './testSetup';
 import * as puppeteer from 'puppeteer';
 import { expect } from 'chai';
 import { killAllChrome } from '../testUtils';
-import { IAttachRequestArgs } from 'vscode-chrome-debug-core';
+import { IAttachRequestArgs, utils } from 'vscode-chrome-debug-core';
 import { getDebugAdapterLogFilePath } from './utils/logging';
+import { DebugClient } from 'vscode-debugadapter-testsupport';
```
digeff: It might be better to rebase the tests from v1. If you add them we are going to have more conflicts later...

src/errors.ts
```diff
@@ -18,3 +18,20 @@ export function getNotExistErrorResponse(attribute: string, path: string): Promi
...
+        id: 2008,
+        format: localize('random.port.no.userdatadir', 'When the remote debugging port is set to 0, you must also provide the "userDataDir" launch argument'),
+        sendTelemetry: true
+    }));
+}
+
+export function couldNotConnectToPort(address: string, port: number) {
+    return Promise.reject(new ErrorWithMessage(<DebugProtocol.Message>{
+        id: 2008,
+        format: localize('launch.port.not.open', 'Could not open a connection to Chrome at: {address}:{port}', '{address}', '{port}'),
```
digeff: Are the latest 2 parameters necessary?

test/int/adapter.test.ts
EricCornelson: I think we chatted about this on an earlier PR, but I can't remember the solution we came to. We can't rebase the whole branch because we'd mess up the history for all of V2, and we can't rebase _just_ the tests.

src/errors.ts
EricCornelson: Not sure I understand the question. You mean address and port on the localize call?

test/int/adapter.test.ts
digeff: You are right. When I say re-base I actually mean merge...

src/errors.ts
digeff: What is the difference between:
localize('launch.port.not.open', 'Could not open a connection to Chrome at: {address}:{port}', '{address}', '{port}'),
and
localize('launch.port.not.open', 'Could not open a connection to Chrome at: {address}:{port}') ?

test/int/adapter.test.ts
EricCornelson: done

src/errors.ts
EricCornelson: Got it. Yeah I think they're just duplicated.

roblourens:
Are all these commits supposed to be in here, like https://github.com/microsoft/vscode-chrome-debug/pull/920/commits/9a85cf5e3e49461b4c117b858f28b17543bd28e4?

EricCornelson:
Per @digeff's request, I did a merge from master, so it grabbed every commit since the last time we merged. Not all commits are actually "merged" since we mostly want to do this for the tests, but some are (e.g. the translations).

roblourens:
Ok. A merge might imply to someone looking at it later that all the changes in master are now in v2 which isn't the case. But I think it's ok as long as we're careful later. It's just important to not lose changes when we merge v2 back into master or however it goes.

PR #923, opened 2019-09-04
Update README's stale DevTools Protocol ref link

zv:


zv:
Chrome's documentation is simply broken under FF (although it does work under Chrome). Closing.

PR #925, opened 2019-09-05
Now we can request dynamic html sources while we are paused

digeff:
Before this fix, if we pause the web-page, and tried to request the source of a dynamic html page, the call to Page.getResourceContent was blocked and only finished after we resumed the debuggee. Because getting sources is an syncrhonic operation in Visual Studio, that blocked the UI.

We now call Page.getResourceContent  for each execution context, so the resources are reloaded pre-emtively, and we then avoid this issue.

Requires this changes in core: https://github.com/microsoft/vscode-chrome-debug-core/pull/524

### DIFF
--- a/src/cdtpComponents/cdtpResourceContentGetter.ts
+++ b/src/cdtpComponents/cdtpResourceContentGetter.ts
@@ -1,7 +1,8 @@
 /*---------------------------------------------------------
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
-import { TYPES, inject, injectable, CDTPEnableableDiagnosticsModule, CDTP, CDTPDomainsEnabler } from 'vscode-chrome-debug-core';
+import { TYPES, inject, injectable, CDTPEnableableDiagnosticsModule, CDTP, CDTPDomainsEnabler, IExecutionContextEventsProvider } from 'vscode-chrome-debug-core';
+import * as _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


src/cdtpComponents/cdtpResourceContentGetter.ts
```diff
@@ -14,8 +15,23 @@ export class CDTPResourceContentGetter extends CDTPEnableableDiagnosticsModule<C
         @inject(TYPES.CDTPClient)
         protected _protocolApi: CDTP.ProtocolApi,
         @inject(TYPES.IDomainsEnabler) domainsEnabler: CDTPDomainsEnabler,
+        @inject(TYPES.ExecutionContextEventsProvider) executionContextEventsProvider: IExecutionContextEventsProvider,
     ) {
         super(domainsEnabler);
+
+        executionContextEventsProvider.onExecutionContextCreated(executionContext => {
+            this.enable().then(() =>
+            /* The first time we call getResourceContent it reloads all resources in case any was evicted from the cache. If the debuggee is paused while that happens,
```
roblourens: Strange, is this an issue in v1 too?

digeff: In v1 we have a different implementation. If we have a .html with 3 script tags we show 3 documents (So we don't need this API in v1).

On v2 instead we show a single document which is the .html web-page (So we need this API).

Issue #927, opened 2019-09-09
"out" folder missing after running npm pack


AObuchow:
- VS Code Version: 1.38.0

Steps to reproduce:

1. Navigate to the root directory of the git repo (after cloning)
2. Run `npm pack`
3. The `debugger-for-chrome-4.11.7.tgz` is missing the `out` directory.

This is an issue for projects trying to integrate the Chrome debug adapter, as is the case for Eclipse's Wild Web Developer which uses `npm pack` for [packaging it's dependencies](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/pom.xml). 

I looked into it a bit and it seems that npm is using the `.gitignore` which is ignoring the `out` directory. My suggestion is to add a `.npmignore` to prevent the `out` folder from being ignored when calling `npm pack`.

PR #928, opened 2019-09-09
Add .npmignore #927

AObuchow:
Add .npmignore to prevent `out` folder from being ignored during `npm pack`


roblourens:
Can you explain what this is for?

AObuchow:
@roblourens This commit is to simplify the integration (packaging) of the Chrome debugger into other IDE's that support the Debug Adapter Protocol. 

Currently, I am working on integrating the Chrome debugger into Eclipse.
For Eclipse, the packaging process currently involves downloading the VSCode extension and then using `npm pack` on the Extension directory. 

Unfortunately, `npm pack` ignores everything on the `.gitignore` unless a `.npmignore` is provided. As a result, the files in the `out` directory were getting left out when running `npm pack`.

roblourens:
Have you considered using `vsce pack` instead? That might work better since it will use the same logic that we actually use to publish this to the app store. I am worried that since I am not using this, I could end up breaking you.

mickaelistria:
For all other language servers and debug adapters (coming from VSCode or VSCode extensions), we're using npm which is standard enough and does a good job at de-duplicating dependencies. Like most people our community has basic enough knowledge of npm to make it work, but doesn't have any knowledge of `vsce`, so it'd be extremely easier to the project and its community if we could solely rely on npm.

roblourens:
Ok, I can merge this but know that I don't offer any guarantees about how well this will work with npm.

Also please don't publish it to npm under the `vscode-chrome-debug` name unless it is under an  organization name or something. (I would just rather leave the name free for me to own in case I do decide to publish it to npm some day)

PR #1022, opened 2020-07-01
Fix picker on v3, and use v3 by default

connor4312:
Fixes https://github.com/microsoft/vscode-js-debug/issues/437

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "ui"
   ],
   "engines": {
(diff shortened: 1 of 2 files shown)


roblourens:
Hm... should this be based on vscode version? I don't know if we want to put this out for people on older versions of vscode

roblourens:
I guess we could just do that by bumping the vscode engine version for the extension?

PR #1049, opened 2020-09-19
fix: only add remote prefix for non-v3 launches

connor4312:
Refs https://github.com/microsoft/vscode/issues/102493#issuecomment-695193077


roblourens (approved):
Do you want this released now?

connor4312:
Sure

